PR #9, opened 2018-11-30
Hari/polyextrude

hsitaram:
Added eb2 feature where extruded triangles can be created by the user. The user specifies 2D coordinates for a maximum of 5 triangles (sort of like a coarse STL file); an eb geometry is created by extruding the triangles. I have also converted the bluffbody and oblique shock cases to use this feature in Exec/Tutorials, they are all working fine.


drummerdoc:
Why is there a hard limit on the number of triangles?
How does one specify the plane of the original triangles? (does it always have to be XY, and extrude in Z?)
Why was it necessary to rename all the files to EB2?  (Is this how Ray moved from EB to EB2?).

rgrout (commented):
There is significant overlap in functionality between this and the distance function generation approach that is part of AMReX now. This way is probably faster, but less general.

I don't see any reason not to have this approach in Pele as well as the distance function approach.

rgrout:
> Why was it necessary to rename all the files to EB2? (Is this how Ray moved from EB to EB2?).

Yes, that was mechanics when migrating from EB to EB2. At this point EB is no longer supported so please remove the 'EB2' designation along with the original versions.

hsitaram:
Why is there a hard limit on the number of triangles?

Ans: The compile time templating of IntersectIF and UnionIF needs compile time specification of number of primitive EB IFs. So I set a maximum and user can give variable number of triangles less than the max.

How does one specify the plane of the original triangles? (does it always have to be XY, and extrude in Z?)

Ans: The aim was to give functionality for extruded polygon-like bodies. Some of our test cases like bluffbody and obliqueshock use these. Moving forward that might be good idea to specify generic plane surfaces which is slowly moving towards CAD like geometries.
 
Why was it necessary to rename all the files to EB2? (Is this how Ray moved from EB to EB2?).

Ans: Ray's convention was followed

hsitaram:
I have renamed folders to EB_ in the hari/polyextrude branch. Let me know if anything else needs attention

drummerdoc:
Hari, Didn't get an answer to the "which plane" question.  Trying to make sure I understand what you did and how to use.  Based on the data, it seems like your triangles are in the z=0 plane and are extruded in the z direction.  How much of this is a hard requirement?

hsitaram:
Sorry. The triangles are specified on the XY plane and extruded along Z. This is a hard requirement for this "extrude" feature, but a more general method would be to have triangle with normal direction specified.

PR #14, opened 2018-12-12
Emmanuel/bugfix forcing

emotheau:
fix a mistake that messed up the tiling process of the forcing routine


emotheau:
The bugfix was commited inside the NSCBC developments, and was merged automatically

PR #17, opened 2018-12-19
Fix for MOL I_R issue

rgrout:


rgrout:
This partial fix / hack no longer required because there is something better in place now - closing.

PR #18, opened 2019-01-02
Some Make.PeleC modifications

jrood-nrel:
I think this may be a mistake in `Make.PeleC`. I seem to have needed this change to allow me to compile with .F90 versions of Fuego files. This change doesn't seem to affect the original `.c` versions of Fuego models.

### DIFF
--- a/Exec/Make.PeleC
+++ b/Exec/Make.PeleC
@@ -72,11 +72,11 @@ ifdef Chemistry_Model
   CHEM_HOME = $(PELE_PHYSICS_HOME)/Support/Fuego/Mechanism/Models/$(Chemistry_Model)
   VPATH_LOCATIONS += $(CHEM_HOME)
   Bpack += $(CHEM_HOME)/Make.package
+  Blocs += $(CHEM_HOME)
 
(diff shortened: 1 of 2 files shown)


marient:
Could you point me to the line in Make.PeleC?

jrood-nrel:
https://github.com/AMReX-Combustion/PeleC/pull/18/files#diff-a71070da25fb2420f39990c86e182bf3R75

marient:
So the repetition of 
Blocs += $(CHEM_HOME1) 
on line 80? 
Yes, I should not have touched it, I thought it was a typo I had introduced.

jrood-nrel:
The redundancy of `CHEM_HOME1` was fixed in this commit and `CHEM_HOME` was missing which was also fixed from this commit. `CHEM_HOME` on line 75 looks to be removed in your pull request, which I feel is problematic. Note `CHEM_HOME` and `CHEM_HOME1` are different.

marient:
Maybe I started from the wrong file, I'll make sure I have the correct one next time.

PR #20, opened 2019-01-08
Emmanuel/gc nscbc cleaning

emotheau:
GC-NSCBC routines have been completely reshaped to be clear and consistent.
Now GC-NSCBC is working in  1D, 2D, 3D with corners


rgrout:
Have not yet gone through this in detail, but:
- Travis says this doesn't build when merged into development
- Have you pulled changes from development in to minimize the size of the merge?
- Is the formulation written up somewhere?
- where can I find the test results written up?
- How does this formulation done on passing a flame across the boundary (e.g., figure 5/6 in the Sutherland and Kennedy paper here: http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.597.6444&rep=rep1&type=pdf)

emotheau:
* Added the missing routine as a BC example, now the code is building properly.
* The development branch is already pulled in the branch.
* Theory can be found here: https://arc.aiaa.org/doi/10.2514/1.J055885
* Haven't tried interaction of flame and BCs.

PR #23, opened 2019-02-12
Remove reset_internal_energy and fix volWgtSum to avoid multiplying m…

rgrout:
…f by vfrac. Latter should not impact correctness there, but avoids bugs if it is copied to remake something like volWgtSumMF that is currently removed

### DIFF
--- a/Source/PeleC.H
+++ b/Source/PeleC.H
@@ -433,8 +433,6 @@ public:
     void react_state(amrex::Real time, amrex::Real dt, bool init=false, amrex::MultiFab* A_aux = nullptr);
 #endif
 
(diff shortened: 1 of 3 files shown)


rgrout:
Assigning @drummerdoc to merge this so that he can make sure CCSE stuff still builds with some routines stripped that I do not think are in use.

drummerdoc:
@rgrout  can you look at the build failure.  That seems like something needing fixing first.

rgrout:
Ok, fixed to allow for the non-EB world to build.

rgrout:
I'm going to merge this in to make it easy to run MMS and then back it out if it breaks...

PR #31, opened 2019-03-05
Emmanuel/hybrid ppm weno

emotheau:
This development brings a new Hybrid PPM/WENO method. Basically it is activated by setting the flag pelec.ppm_type to 3. The difference with ppm=1 or 2 is that the reconstruction at cell faces is done with a WENO interpolation. Currently only 5th order WENO-JS and WENO-Z are implemented. The user can choose with the flag pelec.weno_variant. By default the flag is turned to 1 to use WENO-Z, which provide best results.

All details about this new method and a complete performance analysis can be found here:
E. Motheau and J. Wakefield, "Capturing shocks and turbulence spectra in compressible flows. Part 2: A new hybrid PPM/WENO method," Submitted for publication, 2019. [https://arxiv.org/abs/1902.06666]

The documentation about PPM will be created/updated during the docathon 03/11.

This present development should not impact previous results, it is just a new awesome feature that enable to capture turbulent spectra with high-order accuracy, at the cost of 2nd order Godunov.


drummerdoc:
Not verifying all tests are unaffected....

PR #59, opened 2019-03-18
Cvode c merge

AnneFelden:
This branch is compatible with PelePhysics dev after merge #58


drummerdoc:
I believe this has already been merged. Resubmit if I missed something.

AnneFelden:
I added a regtest for drm chemistry, but it is not absolutely necessary !

PR #63, opened 2019-04-05
Add CTest Capability

jrood-nrel:
This PR adds:

- Ability for developers to run a test suite and build several executables with different options in a single `make`
- GoogleTest framework for creating unit tests (can't believe I forgot to list this originally)
- Ability for test suite to be run nightly with results reported to CDash and much easier to test on multiple machines and compilers, including memory testing, code coverage, etc
- Submodules for AMReX and PelePhysics so it's easier to keep track of external repos
- Ability to build faster using Ninja as a makefile generator
- Ability to build with CMake if wanted

This doesn't replace the current build system and I would work to keep CMake up-to-date. Note, when I tried hooking into CMake that already exists in AMReX, it was broken, and it's also messy, so I will see how managing the sources myself goes initially.

My example CDash is at https://my.cdash.org/index.php?project=PeleC that is testing the `ctest` branch nightly on multiple compilers already, with the address sanitizer turned on in Clang. There are probably more things we can do with this as well. This isn't at all finished, but it's at a point that I think it could be merged so we would be able to build new functionality on top this work. For example, I will need to add more tests, and I would like to get the verification suite into this. Which should be possible in a way such that developers could run it themselves.

Anyway, I'm open to discussion about what it will take to get this merged, and if I made any mistakes here since this is a big PR.

(this branch can be deleted if it gets merged)


drummerdoc:
Jon,
You can certainly refuse the request, but it might be nice to add a description somewhere of what the `ctest` stuff does and how to use/modify the configuration you setup for PeleC for use at other sites.  In particular, what/how it adds to the AMReX reg testing stuff.  Also, does this make the PeleRegressionTesting repo somewhat obsolete?  Is there a higher-level we might consider implementing such that there is a "dashboard" (or whatever clever name they've come up with) that could reflect a status of all or a subset of the Pele codes tested this way?  Finally, how does the gitmodules thing work, and can it be used to manage things like PelePhysics and AMReX for LM and C (and the others) simultaneously, or would make redundant clones all over?
Thanks for your patience with these questions!
-M

jrood-nrel:
Marc, all fair questions. I will try to answer each question one by one.

If this has a chance of being merged, I will put effort into documenting it. Some starter documentation on CTest and mainly how I implement it would be great, and then there is pretty good documentation from Kitware for CTest itself. I just didn't want to go all in if this wasn't going to work for any reason.

In regards to the AMReX regression testing suite, I consider this a full replacement. The AMReX test suite is very limited and involves a large overhead for developers to use in my opinion since I've moved it to a few different machines now. Whereas CTest is extremely accessible to developers, although solution tolerances and storing gold files are the biggest challenge.

This makes the PeleRegressionTesting repo obsolete, except for there being a need to have an external repo host the top-level bash testing script I use to generate nightly results. This repo would be good for that purpose. The explanation of this need is nuanced so I won't explain it in detail here. In short, any changes to the test script need to be updated each night before CTest does the PeleC repo update before testing.

In my opinion, if this were to progress, each Pele code would have its own CDash project site. Each CDash project is allowed 10 build lines for free (paying is not expensive and an option, with the act of having someone pay it monthly being the largest overhead). If we wanted a single dashboard for all projects, I would be inclined to create a github pages site that puts each dashboard into a matrix of frames.

In my opinion git submodules are intended for exactly the purpose in which AMReX and PelePhysics are currently being used. By putting them into submodules, we know exactly where they are, avoiding environment variables, and the PeleC repo tracks pointers to the commits in the submodules that we set. This allows us as developers to more likely be on the same commits at any given time. There is a drawback to this, in that during testing, I would probably want to track the latest AMReX/PelePhysics commits, test, and confirm/commit the submodule updates if they pass the tests. I have not implemented this yet.

If submodules were used in PeleLM as well, there would be redundant copies of AMReX/PelePhysics, etc. However, if this approach were to progress, I would probably entertain the idea of a Pele superset repo that has everything as submodules, and this would avoid the redundancy. Maybe then, CTest could even test the entire Pele suite in this way, making the single dashboard possible as well, but I haven't thought about this until now. That would be interesting.

PR #82, opened 2019-05-22
Hari/mol ext src

hsitaram:
This PR adds the previously available capability to add external_sources, forcing and MMS source terms in MOL_advance; this feature is already there in sdc_advance. Since MOL is used for EB, this will allow simulations with EB to have an external/forcing source terms. 

Marc HDF - I have removed your MMS calls in mol_advance. They will be called when sources are updated. I did the EB_MMS case and it fcompare's to 0.


drummerdoc:
Looks reasonable to me.

PR #83, opened 2019-05-23
Test with fextrema

jrood-nrel:
This uses the `fextrema` tool to check for differences in maxima and minima of plot files from test plots with `fextrema` output from known-to-be-correct "gold" files. The gold files with `fextrema` are portable and are placed in the PeleC repo. Tolerances can also be used for looser checks on different configurations of machine/compiler/optimizations, etc.

Currently a WIP because I want to see how Travis handles the gold files and tolerance.


drummerdoc:
Not sure why but all the PR review requests for PeleC are going into my Junk folder.  I haven't seen any of them.  Sorry for failing to respond.

PR #92, opened 2019-06-06
Consistent Makefiles

jrood-nrel:
This is my attempt and making all the makefiles in Exec consistent. It removes any use of `PELE_HOME` in favor of `PELEC_HOME` and uses `PELEC_HOME` in each `GNUmakefile` to find `Make.PeleC`. If `AMREX_HOME` and `PELE_PHYSICS_HOME` are defined, it should use them as defined. Otherwise, if you don't want to set them, this should still default to the provided submodules. I may be missing why `PELE_HOME` existed, but I assume it was intended to have `PeleC`, `PeleLM`, and `PelePhysics` cloned in the same directory, but in my opinion it made things confusing and seems unnecessary.

There are also makefiles that exist in PelePhysics `Exec` directory, that look like they circle back to PeleC for some reason, but I won't perturb them. Currently it looks like they set up `PELEC_HOME` correctly assuming it's been cloned in `PELE_HOME`.

### DIFF
--- a/Exec/Make.PeleC
+++ b/Exec/Make.PeleC
@@ -1,7 +1,6 @@
-PELE_HOME         ?= ../
-PELEC_HOME        ?= $(PELE_HOME)
-PELE_PHYSICS_HOME ?= $(PELE_HOME)/Submodules/PelePhysics
-AMREX_HOME        ?= $(PELE_HOME)/Submodules/AMReX
+PELEC_HOME        ?= ../
+PELE_PHYSICS_HOME ?= $(PELEC_HOME)/Submodules/PelePhysics
+AMREX_HOME        ?= $(PELEC_HOME)/Submodules/AMReX
 
 TOP := $(PELEC_HOME)
 
--- a/Exec/RegTests/EB_MMS/GNUmakefile
+++ b/Exec/RegTests/EB_MMS/GNUmakefile
@@ -11,9 +11,6 @@ COMP	   = gcc
 USE_MPI    = FALSE
 USE_OMP    = FALSE
 
-# define the location of the PELE top directory
-PELE_HOME    := ../../..
-
 # This sets the EOS directory in $(PELE_PHYSICS_HOME)/Eos
 Eos_dir     := GammaLaw
 
@@ -34,7 +31,9 @@ endif
 
 Bpack   := ./Make.package
 
-include ../../Make.PeleC
+# define the location of the PELE top directory
+PELEC_HOME    := ../../..
+include $(PELEC_HOME)/Exec/Make.PeleC
 
 ifeq ($(DEBUG), TRUE)
 DEFINES += -DDEBUG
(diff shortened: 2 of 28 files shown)


rgrout:
Yes, that was about why PELE_HOME existed, but I agree it’s just confusing now. I’m going to merge this now (before checks complete!) so that we can make sure the “old skool” regression tests build correctly tonight.

drummerdoc:
It looks like PELEC_HOME is only used to to get the constants folder.  Do you guys actually use the constants at all?  If so, should you? I can't imagine that the file is useful for anything.

jrood-nrel:
Thanks! Merging this since Ray asked me to on Slack.

rgrout:
@drummerdoc - occasionally we do out of tree builds, where there isn’t a relative path between the relevant exec directory and the PeleC src

drummerdoc:
@rgrout In that case, you then have to know how to get to the Submodules from the PELEC_HOME folder, I guess.  That seems straightforward.  Maybe there's hidden tricks that come up though?

jrood-nrel:
I moved `PMF` out of the PeleC directory and did this:
```
$ pwd
/Users/jrood/PMF

$ tail GNUmakefile 
# This sets the transport directory in $(PELE_PHYSICS_HOME)/Transport
Transport_dir := EGLib
#Transport_dir := Simple

Bpack   := ./Make.package
Blocs   := .

# define the location of the PELE top directory
PELEC_HOME    := /Users/jrood/combustion/PeleC
include $(PELEC_HOME)/Exec/Make.PeleC
```
and it works.

In regards to `PelePhysics/Testing/Exec`, I don't know what to do about that besides currently `PELE_HOME` or `PELEC_HOME` would need to be set by the user. I won't assume it's being used in `PeleC/Submodules/PelePhysics` to make it default to a relative path to PeleC without them being set. However I have never built anything in PelePhysics directly.

I would still say in the future I think it would be good to have a superset Pele repo that tracks the entire suite as submodules and they could share PelePhysics and AMReX at that level. Then that organization could be implied and solve any issue with how to reference the other projects.

drummerdoc:
Actually, it might be also useful for LM and C to have their own submodule, in case changes to one or more of the dependent repos breaks either in a way that takes some time to fix.

Again, I think the entire reason PELEC_HOME is used in PelePhysics is to bring in a constants file, which contains a bunch of numbers none of us ever use.  If you guys are up for it, it might be useful to strip out the constants you actually need, and put them in a module somewhere and then delete the `constants` folder from PeleC entirely.  I can't imagine that any of the code in `PelePhysics/Exec` actually needs any of those constants.

jrood-nrel:
I agree. That's what I'm proposing: `Pele={PeleC,PeleLM,AMReX,PelePhysics,Etc}` which is a repo that only tracks the submodules. Because in the past it seemed like there was intent to be able to restart simulations between PeleLM and PeleC.

I will put the removal of `PELEC_HOME` in PelePhysics on my list for next week and look at removing the constants directory.

drummerdoc:
Cool.  Given the separation of tasks between C and LM though, and the evolving nature of PelePhysics, I think that we still need to maintain the ability to have LM and C specify their own favorite version of Physics.

PR #93, opened 2019-06-13
Fix intel warnings

jrood-nrel:
This fixes some warnings with the Intel compiler and suppresses some diagnostics from the Intel compiler for CTest.

`regd` isn't being used in `Source/Src_nd/riemann_util.f90` so I set it to `intent(in)` which someone will quickly see that if they need to use it in the future.

I'm also updating the PelePhysics submodule here to utilize a warning fix that was pushed to the repo.


jrood-nrel:
Also, Intel complains about the blank `flush()` statement and it looks unnecessary to me since the code aborts immediately after.

PR #97, opened 2019-06-18
Remove OMP parallel from diffextrap calculation

jrood-nrel:
This also updates the AMReX submodule to use latest development branch which has an OMP data race fix as well.

So it looks like the `diffextrap` calls require a `validbox` to operate on the boundaries correctly but in doing so, it doesn't seem like we can use `omp parallel` here because each thread must operate on the `validbox` which causes race conditions. Even though each thread updates with the same values, it's unnecessary and shows up as data races in the thread sanitizer.

I am open to discussion about a better way to do this. This could probably be done with OMP using some clever logic to only update each thread's piece of the boundary, but the logic in `diffextrap` is already pretty wild. So this would at least correct the behavior for now.

### DIFF
--- a/Source/PeleC_F.H
+++ b/Source/PeleC_F.H
@@ -406,6 +406,7 @@ extern "C"
 
   void pc_diffextrap
     (const int* lo, const int* hi,
+     const int* vlo, const int* vhi,
      BL_FORT_FAB_ARG_3D(TempDiffTerm),
      const int* nc);
 
--- a/Source/PeleC_MOL.cpp
+++ b/Source/PeleC_MOL.cpp
@@ -618,6 +618,9 @@ PeleC::getMOLSrcTerm(const amrex::MultiFab& S,
 
   // Extrapolate to ghost cells
   if (MOLSrcTerm.nGrow() > 0) {
+#ifdef _OPENMP
+#pragma omp parallel
+#endif
     for (MFIter mfi(MOLSrcTerm, hydro_tile_size); mfi.isValid(); ++mfi) {
       BL_PROFILE("PeleC::diffextrap calls");
 
--- a/Source/PeleC_MOL.cpp
+++ b/Source/PeleC_MOL.cpp
@@ -618,13 +618,10 @@ PeleC::getMOLSrcTerm(const amrex::MultiFab& S,
 
   // Extrapolate to ghost cells
   if (MOLSrcTerm.nGrow() > 0) {
-#ifdef _OPENMP
-#pragma omp parallel
-#endif
     for (MFIter mfi(MOLSrcTerm, hydro_tile_size); mfi.isValid(); ++mfi) {
       BL_PROFILE("PeleC::diffextrap calls");
 
-      const Box& bx = mfi.tilebox();
+      const Box& bx = mfi.validbox();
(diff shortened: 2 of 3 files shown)


jrood-nrel:
I have to think about this some more.

jrood-nrel:
Ok reopening this. I think the actual solution here would be to use some amrex calls to fill the ghost cells instead of all this Fortran. However, I think this PR is the temporary fix to at least make this correct when using threading.

jrood-nrel:
Below is and example of how I stepped through it. If you are using a `validbox`, each thread is doing the exact same operations. If you are a `tilebox` each thread races at edges/corners which are 1 cell into another threads ghost cell plane in my example below. This is the example of the left side, and one of the `j` edges. You can clearly see the overlap in threads writing to another thread's ghost cells in both cases.
```
!validbox
!thread 1 line 59
!dif(-1,0:15,80:119,n) = dif(0,0:15,80:119,n)
!thread 1 line 160
!dif(-1,0:15,120,n) = dif(0,0:15,119,n)

!validbox
!thread 2 line 59
!dif(-1,0:15,80:119,n) = dif(0,0:15,80:119,n)
!thread 2 line 160
!dif(-1,0:15,120,n) = dif(0,0:15,119,n)

!tilebox
!thread 1 line 59
!dif(-1,0:15,100:119,n) = dif(0,0:15,100:119,n)
!thread 1 line 160
!dif(-1,0:15,120,n) = dif(0,0:15,119,n)

!tilebox
!thread 2 line 59
!dif(-1,0:15,80:99,n) = dif(0,0:15,80:99,n)
!thread 2 line 160
!dif(-1,0:15,100,n) = dif(0,0:15,99,n)
```

jrood-nrel:
I have also not verified all the `dif` update lines in `Diffusion_nd.f90`, but I'm not convinced there isn't a bug in there as well.

jrood-nrel:
`dg` is 1 for each dimension that exists it seems. It's also unnecessarily multiplied by 1 everywhere.

drummerdoc:
I'm gonna toss that and write this in a way that I understand...hold...sitting in airport terminal typing.

jrood-nrel:
Feel free to fix it however. I have no problem closing this PR with a better solution.

drummerdoc:
How 'bout I push a fix to your branch and you test it in your setup to make sure I didn't make a dumb mistake?

jrood-nrel:
Sounds good to me.

drummerdoc:
Pushed something that compiles, but stepping on the plane.  Can you test, Jon?

jrood-nrel:
Thanks for the fix. I added the `omp parallel` statement back into this and I'm not seeing any data races with it. If we merge it we can see how the tests react to the change. I haven't verified any results.

So with this fix https://github.com/AMReX-Codes/amrex/issues/502 and my OMP pull request, we just need https://github.com/AMReX-Combustion/PelePhysics/pull/27/files merged and PeleC PMF case looks free of data races with threading. I plan to hook the thread sanitizer up to all the tests sometime soon as well.

drummerdoc:
I hope this is all ok.  I assumed you'd be testing this, Jon :)

jrood-nrel:
It looks like the tests were fine after this change. I am not testing any OMP cases, but I will in the future.

PR #99, opened 2019-06-24
Hari/explicit react

hsitaram:
There was a bug in the timestep adaption in the explicit RK chemistry integration. I was not setting the error register to 0 before every time step; hence it was always running at the lowest possible timestep. I also updated sub-modules after which CMake is not working anymore - tried to remove the files that got deleted from amrex, but still having problems.

### DIFF
--- a/CMake/amrex_sources.cmake
+++ b/CMake/amrex_sources.cmake
@@ -27,7 +27,6 @@ function(get_amrex_sources)
    )
    set(AMREX_SOURCE_DIR "${CMAKE_SOURCE_DIR}/Submodules/AMReX/Src/Base")
    add_sources(GlobalSourceList
-     # Utility classes ---------------------------------------------------------
      ${AMREX_SOURCE_DIR}/AMReX_ccse-mpi.H
      ${AMREX_SOURCE_DIR}/AMReX_Array.H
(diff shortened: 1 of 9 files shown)


drummerdoc (commented):
Good in concept...letting Steven weigh in on details.

ghost (approved):
Looks good to me.

jrood-nrel:
I took a look at this. The problems with the CI were due to a CPP file that was just in an inconsistent state. I updated to the latest AMReX to fix it. The constant changes to AMReX source file names is exhausting though. I also fixed the HIT case errors I'm seeing on the tests with this update as well.

drummerdoc (approved):

jrood-nrel (approved):

hsitaram:
Thank you Jon. I wasn't sure what was going on. I did add the new file names to amrex_sources.cmake but it was still failing. should we shift to amrex's cmake. It seems up-to-date. Also, can this be merged?

jrood-nrel:
This can be merged. The filenames were right, but the syntax in one of the CPP files was in a bad state in the AMReX commit at the time of this PR so it was AMReX's fault. I still need to think about how we would be able to use AMReX's CMake directly.

PR #103, opened 2019-07-10
Enabling problem specific tagging.

marchdf:


marchdf:
So as such it doesn't do much. Except that in PeleC, we have a file called `Src_nd/problem_tagging_nd.F90` that defines a basically empty `set_problem_tags` and that a user can overwrite in their run directory with their own problem specific tagging criteria (by supplying their own `problem_tagging_nd.F90`). These couple of lines just calls the `set_problem_tags` defined in `problem_tagging_nd.F90`. And `set_problem_tags`is empty unless the user did something in their run directory to defined it. 

I guess this is just a capability that PeleC _almost_ had, it was just missing these couple of lines to enable it.

drummerdoc:
See if this looks useful:
https://pelelm.readthedocs.io/en/latest/ProblemSetup.html#refinement-criteria

drummerdoc:
If so, I meant to implement a slightly better API for it so as to avoid potential namespace clashing.

marchdf:
So that's a cool capability! And I definitely think it should make it's way into PeleC. But what if I wanted to tag on some weird combination of the state variable and/or coordinate space? This gives the user the ability to hack their own tagging function without messing with the defaults.

drummerdoc:
Can do that...need to make a derived variable though

marchdf:
Two follow-ups on that PeleLM tagging thing you pointed too:
1. Can you specify a region of the domain where refinement is active _and_ turned off in other regions?
2. How soon can we get the PeleLM tagging thing into PeleC? I have an intern working on stuff where it may be useful, and I am happy to help make it happen if you think it is in the right state to pull into PeleC.

drummerdoc:
Could easily add the region (as long as it's a box).  The code I added to PeleLM is in a single commit that I can point you to.  If I were responsible, I'd add it to AmrLevel so you'd have it automatically....I'm not and I didn't.

drummerdoc:
It's actually spread over a few commits.  The first one added tagging by values and gradients, then Emmanuel added the vorticity tagging thing that John likes, then I added tagging rectangles (independent of the state).  The change I'd make is discussed in PeleLM PR#15: https://github.com/AMReX-Combustion/PeleLM/pull/15

rgrout:
@drummerdoc How about we merge this for now, and then move stuff from PeleLM oven if (when?) we have time - but for now this hack is needed to get some production cases we're running to go and I'd like to have those cases based off of a clean development... ok?

drummerdoc:
Fine by me.  I have been using the inputs-file version, and it's pretty decent, so we should move it over someday soon.

PR #105, opened 2019-07-12
Hari/acc prototype

hsitaram:
I have removed eos calls in pc_react_state_expl and also added some dummy acc pragmas


jrood-nrel:
Hari thanks for doing this. I think I will close this for now. We might need this in the future though depending on how the milestone progresses and I start trying to clean up my GPU code along the way

PR #109, opened 2019-08-02
Fix uninitialized arrays

emotheau:
Does this fix the problem of NaNs in NSCBC ?


jrood-nrel:
Yes I think this looks good. Thanks.

PR #110, opened 2019-08-05
Explicit filtering of hydro fluxes and source term.

marchdf:
This commit implements explicit filtering of the hydro fluxes and
source term (for both MOL and PPM code paths). This is necessary to
perform explicitly filtered Large Eddy Simulations.

Contains:
- regressions tests
- documentation
- CMake implementation


marchdf:
Looks like something is broken where I didn't expect it to be. Will reopen once fixed.

PR #111, opened 2019-08-06
Add some missing implicit nones

jrood-nrel:
I noticed some inconsistencies between compilers which seemed to be because of some missing `implicit none`s. So I tried to catch them all here.

### DIFF
--- a/Exec/RegTests/PMF/probdata.f90
+++ b/Exec/RegTests/PMF/probdata.f90
@@ -36,6 +36,7 @@ contains
     use chemistry_module, only : nspecies, get_species_index
     use eos_type_module
 
+    implicit none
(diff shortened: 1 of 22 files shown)


jrood-nrel:
Ok I think this is done now that PelePhysics has received the equivalent changes.

jrood-nrel:
Oh, I didn't expect this to fail on the diffs after updating PelePhysics. I will need to figure this out first.

jrood-nrel:
Oh right, those tg tests that are failing are new from @marchdf and I need to create gold files for those. Back open!

jrood-nrel:
I'm going to merge this since it's been approved and the two tests failing are expected.

PR #112, opened 2019-08-06
Explicit filtering of hydro fluxes and source term.

marchdf:
This commit implements explicit filtering of the hydro fluxes and
source term (for both MOL and PPM code paths). This is necessary to
perform explicitly filtered Large Eddy Simulations.

Contains:
- regressions tests
- documentation
- CMake implementation


jrood-nrel:
It looks like `init_filter` is always called regardless of specifying `pelec.use_explicit_filter`. Is this intended or does it have side effects?

marchdf:
The two failing tests are new tests that I added. @jrood-nrel will add gold files for those using the right machines.

PR #113, opened 2019-08-08
Source terms for LES

marchdf:
This PR implements two LES models for PeleC (constant and dynamic Smagorinsky). The constant Smagorinsky model was verified with MMS. 

It contains:
- LES source term files
- Documentation
- Regression test
- Verification tests

All regression tests pass, except the ones add in this PR and the filtering PR. @jrood-nrel will update the gold files for our machines later. All verification tests pass.


jrood-nrel:
I just have two comments. Do you want to commit that png file, and other stuff in `refdata`? Also, I would change the Peregrine references to Eagle.

marchdf:
I do want that png and stuff in `refdata`. This is reference data that I feel should be part of TG. Good catch on Eagle vs Peregrine.

PR #118, opened 2019-08-30
More fixes to EB OMP data race issue

jrood-nrel:
Updates to the latest AMReX to fix data races in EB during initialization. Puts OpenMP atomics from 3D into 2D as well. Adds ifdefs for OpenMP to hopefully avoid warnings with the Intel compiler. Updates PelePhysics submodule to the latest, and somehow modifications along the way just barely changed the gold file for the fiab-3d test.


jrood-nrel:
These changes give me a clean TSAN report with the piston bowl EB case btw.

PR #122, opened 2019-09-18
Update call to uniform particle initialization

hnkolla:
Implemented Particles Init routine on PeleC-MP side

### DIFF
--- a/Source/PeleCParticles.cpp
+++ b/Source/PeleCParticles.cpp
@@ -272,9 +272,10 @@ PeleC::init_particles ()
         {
             // Initialize uniform particle distribution
             //  {vel(DIM), temperature, diameter, density}
-            ParticleInitType<5, 0, 0, 0> pdata = {{0., 0., 300, 1e-2, 1.}, {}, {}, {}}; //2D
+            //ParticleInitType<5, 0, 0, 0> pdata = {{0., 0., 300, 1e-2, 1.}, {}, {}, {}}; //2D
             //ParticleInitType<6, 0, 0, 0> pdata = {{0., 0., 0., 298, 0.016, 0.68141}, {}, {}, {}}; //3D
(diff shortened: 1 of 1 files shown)


drummerdoc (commented):
Waiting for builds to pass, but a little worried that I can't quite pick up where the separation is between PeleC and PeleC-MP.  Is this documented somewhere (sorry, I should know this...)

hnkolla:
I simply meant that all functionality having to do with Particles class, in this case the SprayParticlesContainer class, is implemented on the PeleC-MP side. I'd made some changes to PeleC-MP, adding a routine to initialize a uniform number of particles, was just having it called from within PeleCParticles.cpp.

drummerdoc:
I should have been more clear in what I was asking.  I noticed that there are PeleC input flags to set particle locations, for example, and that's all done in PeleC-MP.  I was curious then if all controls for MP are passed through PeleC, or if there is a logic behind which are and which dealt with by PeleC-MP directly.  That is, is MP just a helper class with a bunch of functions that are really to be thought of as part of PeleC?  And is this documented anywhere so that a new user understands where to look for things?

hnkolla:
"_Is MP just a helper class with a bunch of functions that are really to be thought of as part of PeleC_?" Yes, that's a fairly accurate description. PeleC-MP, as I see it currently is a part of PeleC, and provides implementations for functions that are called as part of PeleC if PARTICLES are on. Particle specific input params are passed through PeleC. This is not documented, currently, and I'd like to work with Landon to get some basic description in the README.md of PeleC-MP.

drummerdoc:
Sounds good.  Thanks for the info.
BTW: The Travis pipeline passed while checking the PR, but failed with some gpg error when I merged to dev.  I'm not yet sure why, and have no idea if it will affect you.  Let me know if things are as expected.

hnkolla:
Ok, thanks for the heads up. Am testing the build right now, will let you know if something is amiss.

PR #125, opened 2019-09-27
Fix issues with specific configurations

jrood-nrel:
I must have made a mistake when reorganizing the makefiles and this reenables building with DIM=2 with EB. It also explicitly states that MOL without EB is currently unsupported, although we could support it with a little bit of work like I have done in the GPU branch.


drummerdoc:
...so, this means that EB works in 2D now as long as you use MOL?

jrood-nrel:
Yeah

PR #128, opened 2019-10-17
need to use absolute value for velocity component for tagging

hsitaram:
This one line/word change makes sure the absolute value of velocity component is used for tagging, else regions with high negative velocities are not tagged.


rgrout:
Can you separate out the submodule update from the tagging change?

jrood-nrel:
I'm going to close this because I was polluting it with submodule changes. Hari can you create a pull request with just your change?

PR #131, opened 2019-11-15
Merge cvode

EnnaDelfen:
I don't know why this is not there already. This would make cvode avail as an option, instead of dvode, if implicit integration is required. Modifs to PeleC are minor, all calls go through PelePhysics. Everything stays in Fortran for now even though cvode is in cpp.

### DIFF
--- a/Exec/Make.PeleC
+++ b/Exec/Make.PeleC
@@ -66,6 +66,9 @@ else
 endif
 
 # Reactions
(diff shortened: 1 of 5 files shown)


EnnaDelfen:
Yes you  need sundials for cvode, but it should be easy to acquire and build via the ThirdParty folder in PelePhysics. There is a script that git clone sundials and builds everything automatically (or at least it should !).
I can probably change that ifndef in ifdef indeed

PR #134, opened 2019-11-20
Bugfix/unit tests nscbc

asmunder:
These are just very simple fixes to update these test cases so they compile with the latest PeleC. 
I guess merging them should be straight-forward.

Are these tests run in Travis somewhere?


drummerdoc:
@asmunder Thanks for going through those and cleaning them up!

PR #138, opened 2020-01-03
Fix typos in nscbc 3d transverse terms

ghost:
Noticed two typos in the transverse terms for 3D NSCBC. 
This is how I propose to fix them @emotheau 

- T. Bridel-Bertomeu


emotheau:
Thank you very much !

PR #142, opened 2020-01-30
cleaning up explicit react

hsitaram:
Simplifying the logic in explicit react making it more readable. I also removed using eos for cv calculation. Experiments using a pelephysics stand-alone solver suggests this logic is faster.


jrood-nrel (approved):
Sorry Travis CI is still failing on diffs. I do not consider this to be failing. I cannot reproduce the diffs locally, so I'm trying to understand how I can even update the gold files. That shouldn't stop merging this though.

rgrout (approved):
Looks good to me. Thx.

PR #155, opened 2020-07-15
Grown box for the MOL hydro fluxes was incorrect wrt F90 version

marchdf:
This fix ensures that the MOL hydro fluxes are computed on the same
extents that were used in the F90 version of the code. The replaced
code was computing on a box one cell short in the transverse
directions. This shortens the box by one only in the direction
currently considered for the fluxes.

Bug and fix found with help from @nataraj2


jrood-nrel (approved):
Awesome!

PR #162, opened 2020-08-05
Fix memory leak in ckinit

jrood-nrel:
`CKINIT` was being called twice, once with `EOS::init` and also in `reactor_init` when reactions are enabled. This was causing a memory leak since the `TB*` arrays in `mechanism.cpp` were being allocated twice without being freed. This removes the `CKINIT` and `CKFINALIZE` calls from the reactor init and close and only uses the EOS init and close routines. I also removed some unnecessary function wrappers and commented out code that was used for fortran initializations that have gone away.

### DIFF
--- a/SourceCpp/Setup.cpp
+++ b/SourceCpp/Setup.cpp
@@ -91,7 +91,6 @@ set_z_vel_bc(amrex::BCRec& bc, const amrex::BCRec& phys_bc)
 void
 PeleC::variableSetUp()
 {
-
   // PeleC::variableSetUp is called in the constructor of Amr.cpp, so
   // it should get called every time we start or restart a job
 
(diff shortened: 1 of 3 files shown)


SourceCpp/Setup.cpp
```diff
@@ -129,20 +128,14 @@ PeleC::variableSetUp()
...
-
 #ifdef PELEC_USE_REACTIONS
   // Initialize the reactor
   if (do_react == 1) {
     init_reactor();
   }
 #endif
 
-  init_eos();
+  EOS::init();
```
hsitaram: it will be safer to call EOS::init before init_reactor. It may not break anything now but things. in init_reactor may need the chemistry to be initialized before.

hsitaram (approved):

PR #163, opened 2020-08-06
Bugfix: Updating Exec/RegTest/PMF_NSCBC to work with latest PelePhysics

asmunder:
Small bugfix for the Exec/RegTest/PMF_NSCBC case to work with latest PelePhysics.
I literally just copy-pasted the files `pmf_generic.f90` and `probdata.f90` from Exec/RegTest/PMF, where this was already fixed.

### DIFF
--- a/Exec/RegTests/PMF_NSCBC/pmf_generic.f90
+++ b/Exec/RegTests/PMF_NSCBC/pmf_generic.f90
@@ -9,7 +9,6 @@ module pmf_module
 contains
 
   subroutine read_pmf()
-    implicit none
     character (len=20) :: ctmp1, ctmp2, fmt
     character (len=1) :: ctmp
     integer index, found, ist, reason, i, j, lsize, pos1, pos2, n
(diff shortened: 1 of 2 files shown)


jrood-nrel (commented):
Is there a reason for removing `implicit none` in these files?

asmunder:
> 
> 
> Is there a reason for removing `implicit none` in these files?

The only reason is that it was done in the Exec/RegTest/PMF files that I copied from. I haven't checked if that changed in the PMF case recently, or whether this difference has been there for a long time. 

I have no strong opinion, I can put it back if you wish. Perhaps also in PMF case then, for consistency? (I don't know if it's possible/desireable to have these common files in a common location, or not.)

jrood-nrel:
`implicit none` should be kept. The PMF case should add them if they are not there, but let's just merge this PR as long as they aren't removed here.

jrood-nrel:
@asmunder Thanks for finding this. I made these changes without removing the `implicit none`s and I will now close this.

PR #177, opened 2020-10-16
EBVerification Case C2

nataraj2:
Adding a quiescent flow test case with complex geometry - union of a sphere and cylinder with species.


jrood-nrel (changes requested):
Can we remove the `datlog`, `movie.visit`, `submit.batch`, and the `visitlog.py`?

marchdf:
Hey @nataraj2 , a couple comments:

- do you need both `pp.py` and `postprocess.py`?
- could you add the relevant documentation and figures into the RTD sphinx documentation? You can follow the examples for the other cases there.

jrood-nrel:
I think this is out of date. I'm going to close it. Please reopen a new PR if you would like to add EB-C2.

PR #180, opened 2020-10-22
Free arrays allocated in React.cpp after use to solve memory leak

jrood-nrel:
I'm a little confused. It seems like this memory leak should be causing problems for anything using reactions and it's allocation sizes are pretty major. I think I'm still missing something here because I was only seeing this when using SUNDIALS.

I also did a `clang-format` on this file.

### DIFF
--- a/SourceCpp/React.cpp
+++ b/SourceCpp/React.cpp
@@ -96,15 +96,17 @@ PeleC::react_state(
       const int do_update =
         react_init ? 0 : 1; // TODO: Update here? Or just get reaction source?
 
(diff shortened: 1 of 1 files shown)


hsitaram:
I can't believe I missed those cudafrees. I was doing it in pelephysics test cases. Those frees are only needed for sundials integrators. hope that clears your confusion

jrood-nrel:
Hah, oh right, these changes are all under one big `#ifdef USE_SUNDIALS_PP`. I think everything I was seeing makes sense to me now.

hsitaram (approved):

marchdf (approved):

PR #182, opened 2020-11-06
Hari/refactor

hsitaram:
This PR mainly does some refactoring
1) Mol advance seems to use S for state and source interchangeably which makes it difficult to read.
moving forward I hope we can stick to S/Snew/Sold for state and not sources
2) I have also added comments in the react section on what each section of code is doing. I have removed the use of "a" as the source term, instead called it non-reacting source (nr_src)
3) While doing this, i found out that we are not updating the total energy at the end of reactions, which is also fixed now

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,9 @@ name: PeleC-CI
 
 on:
   push:
-    branches: [development]
+    branches: [hari/refactor]
   pull_request:
-    branches: [development]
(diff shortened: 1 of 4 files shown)


jrood-nrel (approved):
Looks fine with me. Thanks for doing this.

marchdf:
Love this kind of thing.

How big of a problem do you think this `we are not updating the total energy at the end of reactions` was?

marchdf (approved):

PR #183, opened 2020-11-12
EB Verification: Cases C4 C5 and Bug Fix

baperry2:
This pull request adds EB Verification cases C4 and C5 (temperature diffusion). 

It also fixes a bug for the EB treatment in Diffusion.cpp that was causing different results when C5 was run using different numbers of processors. The issue comes up when there is a FAB that has cut cells only in the ghost cells but not in the interior cells. Before doing EB stuff, there are checks to see whether the FAB has cut cells, but in the CPP version of the code this check is over vbox (no ghost cells) instead of cbox (one ghost cell) as in F90. At least for pc_fix_div_and_redistribute, we need to deal with cut cells in the ghost region, so replacing with cbox fixes the issue.


SourceCpp/Diffusion.cpp
```diff
@@ -136,6 +136,10 @@ PeleC::getMOLSrcTerm(
         }
         continue;
       }
+      // Note on typ: if interior cells (vbox) are all covered, no need to
```
marchdf, line 139: So why not making `typ` above be on `cbox` and not `vbox` that way it is not redefined. Also it would match F90 implementation better.

marchdf (changes requested):

SourceCpp/Diffusion.cpp
baperry2: That's the way I had it originally. The downside is that if you have a FAB that is all covered cells on the interior but has cut cells in the ghost region you would end up going through the whole getMOLSrcTerm calculation for no reason. I don't think this would actually hurt anything, but it would be slightly inefficient in the cases where it comes up.

In the end I don't think it makes much difference, so I'm happy to switch it back to the more simple way if you prefer it.

jrood-nrel:
Looks good. I love to see fixes. I'm fine with whatever @marchdf recommends for the `Diffusion.cpp` change.

jrood-nrel (approved):

SourceCpp/Diffusion.cpp
marchdf: got it! I was being dumb

marchdf (approved):

PR #184, opened 2020-11-16
Implementing Bruce's suggestion - remove flux interp at cut faces

nataraj2:
This change removes the flux interpolation at the cut faces by zeroing out the interpolation coefficients. 
The first value will be `fa(i, j, k)` - the cut face area fraction. All other 3 values which involve the interpolation are zeroed out.


jrood-nrel:
I don't think we're merging this so reopen if this is something we need to discuss.

PR #192, opened 2020-12-01
Solve loophole static analysis found where non_react_src could be uninitialized

jrood-nrel:
I think this was a false positive from my analysis and I don't think `non_react_src` could have ever been uninitialized, but I think fusing the `if` condition should solve this. @hsitaram Please verify the change in logic still makes sense.


hsitaram (approved):
looks good! simpler logic.

PR #205, opened 2020-12-21
Pistonbowl fix by clipping/normalizing reaction mass fractions

marchdf:
This PR introduces an optional clipping and renormalizing of the mass fractions that are used to compute the reactions. It does not clip/renormalize the mass fraction state so negative mass fractions will still be observed in the state. However these negative mass fractions will not negatively affect the reaction computations. 

This PR fixes the pistonbowl failures we were observing, arising from the creation of negative mass fractions. The attached [input file](https://github.com/AMReX-Combustion/PeleC/files/5725198/inputs-3d.txt) was used for this simulation (note the increased background temperature to get the flow to burn quicker) and can be used with`pelec.do_clean_react_state=0` or `1` to observe failure or success of the simulation, respectively. The resulting temperature field looks like: 
movie

Negative mass fractions are observed in the flow field and these reach a minimum of `-1e-5` for CH4 and `-1e-6` for HO2 mass fractions.

I am not going to argue that this is the right long term fix. That would require furthering Bruce and Mahesh's work in understanding the specifics of the discretization scheme that lead to negative mass fractions. Specifically, it would require a modification of the discretization scheme to improve scalar boundedness (a thing we know is not part of the discretization specs). However, this fix improves robustness of the code and allows us to proceed with simulations such as the pistonbowl. Happy to engage in discussion on how we want to proceed in the near term/long term.

One thing to note is that this is only implemented for the implicit chem integrator. I can do it for the explicit as well if desired.

Thanks to @hsitaram for suggesting this specific version of the fix. Thanks to @baperry2 and @nataraj2 for trying initial simulations of this case indicating that this may be a way to proceed.

### DIFF
--- a/SourceCpp/Params/_cpp_parameters
+++ b/SourceCpp/Params/_cpp_parameters
@@ -380,7 +380,7 @@ adaptrk_nsubsteps_guess      int           50                 n
 adaptrk_errtol               Real          1e-16              n
 
 #flag to clean the state for the reactions
-do_clean_react_state         int           1                  n
+clean_react_massfrac         int           1                  n
 
 #-----------------------------------------------------------------------------
 # category: parallelization
--- a/SourceCpp/Params/_cpp_parameters
+++ b/SourceCpp/Params/_cpp_parameters
@@ -368,16 +368,19 @@ disable_shock_burning        int           0
 chem_integrator              int           1                  n
 
 #explict RK chemistry integrator options (minimum substeps)
-adaptrk_nsubsteps_min        int          20                  n
+adaptrk_nsubsteps_min        int           20                 n
 
 #explict RK chemistry integrator options (maximum substeps)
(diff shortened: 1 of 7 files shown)


hsitaram (approved):
looks good to me!

drummerdoc (commented):
Note that an alternative to this is the so-called JBB_HACK, where the unhacked states are saved, and the results of the chem integrator are added to the original states.  This preserves mass conservation while also allowing the benefits of the hack.  It's more emotionally appealing, but has questionable real utility...might be worth considering as an alternative.

this actually might be exactly what's this PR does...

hsitaram:
I guess we are doing the "JBB_HACK" without knowing its technical name, :)

jrood-nrel (approved):

shashankNREL (approved):
Looks good to me.

PR #210, opened 2020-12-23
Energy Equation in Explicit Chemistry Integrator

baperry2:
This pull request updates the PeleC explicit chemistry integrator to solve the internal energy equation instead of the temperature equation. @hsitaram mentioned that solving the temperature equation is a historical artifact relating to Fuego having an analytical Jacobian based on temperature, but that is not relevant for the explicit solver. 

For a constant volume reactor, internal energy is the conserved quantity and its equation is simply `d/dt(rho e) = external forcing`. The present implementation recasts this to a temperature equation using `d/dt(rho e) = rho*cv*dT/dt + sum(wdot_i*e_i)`, which assumes ideal gas behavior.

There are a few advantages to switching to the internal energy equation:
- The simpler form of the equation allows for simpler code and eliminates the need for EOS calls to determine cv and e_i. 
- The current implementation tracks two temperatures that nominally should be equivalent. The revised implementation eliminates this as temperature is always obtained from internal energy through an EOS call. 
- The internal energy implementation does not assume ideal gas behavior and therefore can be used for real gas equations of state

I tested the revised code using the zeroD RegTest. The figure below shows that the autoignition calculation is essentially unchanged. However there are small quantitative differences that arise because of numerical differences in explicit integration of `rho*cv*dT` vs. `de`.

image

### DIFF
--- a/SourceCpp/React.H
+++ b/SourceCpp/React.H
@@ -144,7 +144,7 @@ pc_expl_reactions(
   amrex::Real rhoydot_ext[NUM_SPECIES];
   for (int n = 0; n < NUM_SPECIES; n++)
     rhoydot_ext[n] = nr_src(i, j, k, UFS + n);
-
+         
   // RK dts
   amrex::Real dt_rk = dt_react / nsteps_guess;
   const amrex::Real dt_min = dt_react / nsteps_max;
@@ -173,40 +173,26 @@ pc_expl_reactions(
 
     amrex::Real wdot[NUM_SPECIES] = {};
     amrex::Real massfrac[NUM_SPECIES] = {};
-    amrex::Real ei[NUM_SPECIES] = {};
     for (int stage = 0; stage < 6; stage++) {
+      // Find up to date temperature and mass fractions
       rhoInv = 1.0 / urk[URHO];
       for (int n = 0; n < NUM_SPECIES; ++n) {
         massfrac[n] = urk[UFS + n] * rhoInv;
       }
-
-      EOS::RTY2WDOT(urk[URHO], urk[UTEMP], massfrac, wdot);
-
       amrex::Real Temp_rk;
       EOS::EY2T(rhoe_rk / urk[URHO], massfrac, Temp_rk);
 
+      // Compute species source terms
+      EOS::RTY2WDOT(urk[URHO], Temp_rk, massfrac, wdot);
       for (int n = 0; n < NUM_SPECIES; ++n)
         wdot[n] += rhoydot_ext[n];
-
-      // get mixture specific heat
-      amrex::Real cv;
-      EOS::TY2Cv(urk[UTEMP], massfrac, cv);
-
-      // get species internal energies
-      EOS::T2Ei(Temp_rk, ei);
-
-      // update temperature eqn source term
-      amrex::Real tempsrc = rhoedot_ext;
-      for (int n = 0; n < NUM_SPECIES; ++n)
-        tempsrc -= wdot[n] * ei[n];
-      tempsrc /= (urk[URHO] * cv);
 
       /*================== Update urk_err =================== */
       // Species
       for (int n = 0; n < NUM_SPECIES; ++n)
         urk_err[UFS + n] += err_rk64[stage] * dt_rk * wdot[n];
-      // Temperature
-      urk_err[UTEMP] += err_rk64[stage] * dt_rk * tempsrc;
+      // Energy
+      urk_err[UEDEN] += err_rk64[stage] * dt_rk * rhoedot_ext;
 
       /*================== Update Stage solution =================== */
       // Species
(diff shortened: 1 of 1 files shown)


drummerdoc:
The formulation is only tangentially related to the form of the Jacobian - the chain rule is used to translate to whichever variables are necessary to compute the Jacobian, so that is only an argument for reduced complexity.

The real reason to evolve T with the other variables is that it impacted the efficiency of several difficult test cases for PeleLM. In particular, I noticed it seemed to help limit the time step with the implicit CVODE integrator.  We could have formulated it entirely without an energy equation (and in fact, in terms of Y, rhoY, C or X) and with an energy equation based on T, e, or h.  Without rigorous testing, the (rhoY,T) appeared to be most robust across an arbitrary range of problems.  

The Castro folks had the same observation with their system.

It is reasonable to wonder if the changes in this PR have any impact in the efficiency of the solver.  They may not, but it is worth a quick check.

hsitaram:
In light of what Marc said, can you add a parmparse flag to choose this
formulation as well as the default temperature way. This way we can turn it
on/off for non-ideal/ideal gas flows.

marchdf:
@drummerdoc can we get more details on "it impacted the efficiency of several difficult test cases for PeleLM" ? Are there some tests from LM that would make sense to pull into C to see if there's a difference in efficiency?

shashankNREL:
@drummerdoc: I am actually surprised to hear that
 internal energy equation impacted efficiency because it is conserved quantity and should be easier to solve when compared to Temperature which has a source term which has a time scale very different from other terms in the transport equation such as advection and diffusion.

baperry2:
I agree with Shashank. It makes sense to me that temperature would be more stable for an implicit solve. But for the explicit solve in the PR (which doesn't touch any of the PelePhysics sundials stuff), temperature should be more stiff than internal energy, for which the only source term is the external forcing from transport. I did look into the cost of the H2 ignition test case and that supports this point. The new code is faster (47s vs. 56s total wall time), and the speed increase largely comes from taking fewer rk chemistry substeps at each timestep, as shown in this plot:
Screen Shot 2020-12-23 at 8 41 06 AM

I'm actually surprised by the amount of difference. My expectation was that if temperature was causing the adaptive steps to become smaller, then there would be at least one species with the same limitation anyway. But the number of chemistry substeps actually bumps into the maximum (300) for the temperature equation, but stays well below it for the internal energy equation. I'm planning to check a couple other cases to see if this trend is general.

drummerdoc:
1. There are no formal test cases backing up my statement.  These observations were from back around 2000-2001 when I first wrote LMC with DVODE, and maybe specifically a little later when I was working with GRI30 and nitrogen chemistry.
2. T included in the system as an auxiliary variable had the effect of limiting the time step more than otherwise, and probably resulted from the fact that we did not carry around a vector of typical values at that point.  Since all variables were being clamped to the same tolerances, the large T values tended to make the T equation play a more significant role in computing the limiting dt.  This is entirely independent of energy transport, since at least at the time it was in the context of Strang-splitting - which, incidentally made life very difficult for the solver, since diffusion would act independently to continually populate the cells with garbage that DVODE had to clean out before it could proceed with the physical advance.
3. The impact of T on the evolution of the system varied dramatically, depending not only on the chemical model but on the system being advanced.  Diffusion flames acted different from premixed, and both were different than turbulent flames.  So a single test was inconclusive, and a single solution was not optimal for all cases.
4. In the Strang-split context, the ODE being solved was constant rho, constant rhoh.  (thus p drifted during the solve...it was actually pretty weird).  But because of that it was over constrained when T was added.  We would throw away the evolved T and replace it in the fKernel function with the T from the EOS.  it was used purely to impact the local dt for DVODE.
5. Castro used the same strategy, but left the ability in to select between all the various options (a hallmark of that code...to keep every option for every decision allowing an infinite set of runtime combinations...another discussion).  Their case was particularly foul because the reaction rates scale something like T to the 13th power or something awful. They also found that carrying along the T equation helped the robustness of the ODE integration...also in a Strang-split scenario.  They too have moved to an SDC approach, where the rhs for the system includes a source terms for A+D, and so maybe the role of T has changed.  
6. It is still the case that if the system is PROPERLY scaled, an auxiliary equation should have no impact on the evolution, so there's that.  But as I mentioned, it is difficult, if not actually unclear, how to "properly" scale the system.
7. Because our internal energy and enthalpy variables in combustion can be negative, extra care needs to be applied to properly compute "typical values" of these quantities.  I always did something like TV = MAX(h) - MIN(h), where the operators were over the entire domain.  This has funny repercussions though...not at all clear.  T is much nicer in this context, though, as pointed out, presents other difficulties in the case of non-ideal mixtures.

baperry2:
Closing this pull request because pending a larger change to the error scaling the preference seems to be integration based on the temperature equation. The integration based on the energy equation will be kept for non-ideal equations of state only in a forthcoming pull request that contains all changes required for that.

PR #221, opened 2021-01-11
Use thrust for CUDA builds, and custom function for other builds

marchdf:
Non CUDA GPU builds should use the custom functions since thrust may not be available.


jrood-nrel (approved):
Looks like the right move to me.

PR #229, opened 2021-01-20
Turn off FPE trapping for MASA initialization

jrood-nrel:


jrood-nrel:
I am aware I am merging this with the tests failing, but errors aren't being introduced in this PR. They are being exposed.

PR #232, opened 2021-01-25
Remove CUDA-specific reactions code

jrood-nrel:


SourceCpp/React.cpp
```diff
@@ -146,30 +146,26 @@ PeleC::react_state(
...
+          amrex::Vector<amrex::Real> h_rY_src_in(ncells * NUM_SPECIES);
+          amrex::Gpu::DeviceVector<amrex::Real> rY_src_in(ncells * NUM_SPECIES);
+          amrex::Real* d_rY_src_in = rY_src_in.data();
+          amrex::Vector<amrex::Real> h_re_in(ncells);
+          amrex::Gpu::DeviceVector<amrex::Real> re_in(ncells);
+          amrex::Real* d_re_in = re_in.data();
+          amrex::Vector<amrex::Real> h_re_src_in(ncells);
+          amrex::Gpu::DeviceVector<amrex::Real> re_src_in(ncells);
+          amrex::Real* d_re_src_in = re_src_in.data();
+
```
hsitaram, line 162: do we have to clear these at the end like h_rY_in.clear()?

jrood-nrel: This is a good thought, but I think when they fall out of scope, their destructor does it. I will double check.

jrood-nrel: I can't find any example of a `clear` happening in AMR-Wind and they don't mention it here https://amrex-codes.github.io/amrex/docs_html/GPU.html#gpu-vectors so I'm assuming all these vectors get cleared automatically even on device.

hsitaram: I was hoping that will be the case.. you never know sometimes!

hsitaram (approved):

SourceCpp/React.cpp
jrood-nrel: Definitely. I will still need to do more testing of this by running longer simulations once it's merged.

PR #235, opened 2021-02-01
Hari/sundials boxinteg

hsitaram:
Adding chemistry integration for sundials which takes in array4s instead of flattened arrays. It should give better performance on GPU for larger problem sizes. Although, i didn't see much with the regtest/PMF case. We may have to merge Jon's "managed memory removal" PR first and I can merge it into this branch.


marchdf:
This looks fine, but 2 questions: 
- why is PMF CI failing?
- do we need another ifdef or can we do without? I know Jon is trying to minimize those...

hsitaram:
I am looking at why CI is failing. Nothing wrong with CI, I am making a mistake.

jrood-nrel (commented):
I have three comments:

1. Can this be a `pelec.chem_integrator=3` that is under `USE_SUNDIALS_PP` instead of a new ifdef? I can add cvode to the SUNDIALS submodule.
2. I don't think this needs to be CUDA-specific. Can you make it more general like #232 ?
3. I'm guessing something is going out of bounds since the debug build is failing.

hsitaram:
i like the idea of the run-time option for pelec.chem_integrator. Let me first get CI to go through and I will make that change. I have merged with #232, so now using AMREX_USE_GPU

hsitaram:
I have addressed everything - no more AMREX_USE_CUDAs, new pelec.chem_integrator option (number 3), CI should go through now

SourceCpp/React.cpp
```diff
@@ -194,61 +234,84 @@ PeleC::react_state(
...
-              d_re_src_in[offset] = rhoedot_ext;
+                if (captured_clean_react_massfrac == 1) {
+                  clip_normalize_rYarr(i, j, k, sold_arr, rhoY);
+                }
+              }
             });
 
-          amrex::Gpu::streamSynchronize();
+#ifdef AMREX_USE_GPU
+          cuda_status = cudaStreamSynchronize(amrex::Gpu::gpuStream());
```
jrood-nrel: This is still a cuda call.

SourceCpp/React.cpp
```diff
@@ -146,6 +175,14 @@ PeleC::react_state(
           amrex::Real chemintg_cost;
           amrex::Real current_time = 0.0;
 
+#ifdef AMREX_USE_GPU
+          cudaError_t cuda_status = cudaSuccess;
```
jrood-nrel: Can we remove this?

SourceCpp/React.cpp
```diff
@@ -135,7 +162,9 @@ PeleC::react_state(
                 nsubsteps_max, nsubsteps_guess, errtol, do_update,
                 captured_clean_react_massfrac);
             });
-        } else if (chem_integrator == 2) {
+        }
+
+        else if (chem_integrator == 2 or chem_integrator == 3) {
```
jrood-nrel: Prefer `||`

SourceCpp/React.cpp
```diff
@@ -103,6 +128,7 @@ PeleC::react_state(
       const int do_update = react_init ? 0 : 1;
 
       const int captured_clean_react_massfrac = clean_react_massfrac;
+      const int captured_chem_integrator = chem_integrator;
```
jrood-nrel: Can you reduce the scope of this so it doesn't cause an unused variable warning?

jrood-nrel:
Thanks @hsitaram . I made a few more comments, but they're minor. I have one more question. It looks like the PelePhysics submodule would get changed with this merge. Is that necessary?

SourceCpp/React.cpp
hsitaram: removed this

SourceCpp/React.cpp
hsitaram: sure. it is now "||"

SourceCpp/React.cpp
hsitaram: I moved it into the sundials section

jrood-nrel (approved):

marchdf:
@hariswaran I know this was merged, but it feels like we need a CI check and/or a regression test exercising this new capability. Do you think you could add that?

PR #236, opened 2021-02-01
Fix EB C8 plt file with llvm

marchdf:
I think reactions source needs to be reset to zero at the beginning of the step such that, when in debug mode with llvm, NaN's don't get written to the plt file for the `rho_omega_*` components.


jrood-nrel (approved):

hsitaram (approved):
yes, good catch!

PR #241, opened 2021-02-03
Remove managed variables from passive map

jrood-nrel:
This was not a fun one. Nor is the solution great.

@ldowen I also changed the `for` loops using `!=` to `<` to keep the code more consistent. Hope you don't mind.


marchdf:
This looks fine by me. Seems like the CI caught some stuff though.

ldowen:
No problem. I will pull this and check it out real fast.

PR #243, opened 2021-02-05
Fix underoverflow in EB-C8

marchdf:
Apparently EB-C8 was a good one... The fix I thought fixed EB-C8 was definitely a needed fix. But it wasn't the fix needed for the nightly tests. EB-C8 with llvm generates underflow calculations (usually uncaught) and the result of these are written to file. These are then attempted to be read in with `fcompare` and generate a `amrex::Error::0::Read of Vector<Vector<Real>> failed` because it fails to read a number like `-9.5044093114872298e-315`. The way to fix this is to ensure no underflow computations are done in the code. This PR fixes two identified underflow errors in EB-C8:

- an underflow calculation in the riemann solver (fairly straightforward)
- an underflow calculation arising from tiny mass fractions (e.g. around 1e-160). In this case, the fix is to ensure that any mass fraction far below machine precision and close to zero is turned into an actual zero. It is enough to do this in `ctoprim`. 

I should note that I expect this PR to introduce diffs in the nightly tests... But those diffss should all be below machine precision.

### DIFF
--- a/SourceCpp/Riemann.H
+++ b/SourceCpp/Riemann.H
@@ -24,7 +24,7 @@ riemann(
   const amrex::Real spr[NUM_SPECIES],
   const amrex::Real gamcr,
   const int bc_test_val,
-  const amrex::Real csmall,
+  const amrex::Real /*csmall*/,
   const amrex::Real cav,
   amrex::Real& ustar,
   amrex::Real& uflx_rho,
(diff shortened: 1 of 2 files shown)


jrood-nrel (approved):

jrood-nrel:
Nice work finding these.

PR #244, opened 2021-02-05
Landon/soot and spray

ldowen:
I am finally doing a PR for this. There are a few things I am not 100% sure will mess things up. Specifically, this
https://github.com/AMReX-Combustion/PeleC/blob/3221a4c0ea14ee92a0f6c70af94aaaf11e052b0c/SourceCpp/Diffusion.cpp#L126
But I had issues where my Sborder size was bigger than it needed to be for the diffusion stuff and it caused an error. So I had to explicitly set it and found that only 2 ghost cells were needed if EB was turned off. Also, I would like to keep my branch around even after it merges, if possible.


ldowen:
I am not up-to-date on all of the methodology for PRs now. I will update it ASAP.

ExecCpp/Make.PeleC
```diff
@@ -108,16 +108,24 @@ ifeq ($(USE_EB), TRUE)
...
+endif
+
+ifeq ($(USE_SOOT), TRUE)
+# Ensure the number of soot moments is correct
+    ifeq (,$(filter $(NUM_SOOT_MOMENTS),3 6))
+        $(error NUM_SOOT_MOMENTS must be 3 or 6)
+    endif
+  DEFINES+=-DSOOT_MODEL
+  DEFINES+=-DNUM_SOOT_MOMENTS=$(NUM_SOOT_MOMENTS)
+  Bpack   += $(PELEC_MP_HOME)/Source/Soot_Models/Make.package
```
marchdf, line 127: Is PeleC_MP a submodule of PeleC now? Maybe it should be?

ExecCpp/Make.PeleC
```diff
@@ -151,6 +170,21 @@ AMReX_buildInfo.cpp:
 
 include $(AMREX_HOME)/Tools/GNUMake/Make.rules
 
+ifeq ($(USE_SUNDIALS_PP),TRUE)
```
marchdf, line 173: Is this necessary? Don't we do sundials as a submodule now?

Source/PeleC.H
```diff
@@ -50,6 +54,9 @@ enum sources { ext_src = 0,
 #endif
                diff_src,
                les_src,
+#ifdef SOOT_MODEL
+	       soot_src,
```
marchdf, line 58: I guess we have to run `clang-format` on this code. Looks like the CI isn't passing.

Source/PeleC.cpp
```diff
@@ -152,6 +155,10 @@ bool         PeleC::do_mol_load_balance = false;
 SprayParticleContainer* PeleC::SprayPC = nullptr;
 #endif
 
+#ifdef SOOT_MODEL
+SootModel* PeleC::soot_model = new SootModel();
```
marchdf, line 159: is there a way we can avoid raw pointers? `unique_ptr` or whatnot and then avoid all the `delete` and stuff like that?

Source/PeleC_setup.cpp
```diff
@@ -225,9 +225,20 @@ PeleC::variableSetUp ()
 	cnt += NumSpec;
     }
 
+#ifdef SOOT_MODEL
```
marchdf, line 228: Is it necessary to use ifdefs for the `SOOT_MODEL`? Can they be replaced by `if` conditions? This would help in the test matrix of exectuables.

Source/Src_1d/riemann_1d.F90
```diff
@@ -8,7 +8,7 @@ module riemann_module
                                    NGDNV, GDU, GDPRES, &
                                    URHO, UMX, UEDEN, UEINT, &
                                    small_temp, small_dens, small_pres, &
-                                   npassive, upass_map, qpass_map, &
+                                   npassive, npassnm, upass_map, qpass_map, &
```
marchdf, line 11: These are changes to fortran code. Aren't we moving to C++ only?

SourceCpp/Diffusion.cpp
```diff
@@ -117,11 +117,14 @@ PeleC::getMOLSrcTerm(
     // }
     // const int* domain_lo = geom.Domain().loVect();
     // const int* domain_hi = geom.Domain().hiVect();
-
     for (amrex::MFIter mfi(MOLSrcTerm, amrex::TilingIfNotGPU()); mfi.isValid();
          ++mfi) {
       const amrex::Box vbox = mfi.tilebox();
-      int ng = S.nGrow();
+#ifdef PELEC_USE_EB
```
marchdf, line 123: Ok I see what you mean... changing this always causes us headaches. We will have to hope the CI and nightlys catch all the potential memory issues that might arise from this. I am a little confused as to why this is necessary... could you elaborate?

SourceCpp/Params/param_includes/pelec_defaults.H
```diff
@@ -3,17 +3,17 @@
 // or add runtime parameters, please edit _cpp_parameters and then run
 // mk_params.sh
```
marchdf, line 4: There shouldn't be any chances here beyond the soot stuff. And this file should be auto generated through `mk_params.sh`

marchdf (changes requested):
I made a first coarse pass and left some comments.

SourceCpp/Diffusion.cpp
ldowen: For the spray stuff, I Sborder could potentially have as much as 6 ghost cells. However, the diffusion operations are completely independent of the spray operations. So the number of ghost cells should not depend on the size of Sborder. It causes problems trying to solve the derivatives 5-6 ghost cells out if I don't change ng.

ExecCpp/Make.PeleC
drummerdoc: This is an ongoing irritation.  For the foreseeable future, if the Chemistry_Model == Null, sundials should not be needed by any of the Pele codes, otherwise it should be needed by all of them.  I'm thinking that USE_SUNDIALS_PP should be replaced everywhere by PELE_USE_CHEM, and that should be set FALSE if Chemistry_Model==Null, and TRUE otherwise.

Can anyone think of objections to this strategy?

marchdf: I have no objections. I would say that we should see how it is currently done in PeleC (I think this differs from that but I don't tend to deal with that stuff so I may be wrong). And then do what you suggest but in a separate PR...

Source/PeleC.cpp
ldowen: Yep. I will update that.

ExecCpp/Make.PeleC
ldowen: Since I use both PeleC and PeleLM, I don't keep things in submodules. I hadn't realized this stuff differed from the development branch because I added it so long ago. I will remove it for now and we can figure it out separately.

hsitaram: This change has to come also from pelephysics side, its build system has to be updated. There are couple others like CVODE_BOXINTEG that is unnecessary. guess this is a separate discussion from this PR.

drummerdoc: Yes, we'll clean that up in a separate PR for PelePhysics that will then trigger edits in PeleLM and PeleC

Source/PeleC_setup.cpp
ldowen: You mean eliminate the preprocessor directives for soot altogether and change them to C++ checks? Or just change the preprocessor from ifdef to if?

ExecCpp/Make.PeleC
jrood-nrel: It is a submodule.

ldowen:
There are too many problems with this PR. I didn't realize how much the PR methodology had changed. I am going to close it, implement recommendations provided here, and start over.

Source/PeleC_setup.cpp
marchdf: yes, sorry I mean c++ checks

ldowen: I could get rid of it once I clean up the soot code and remove all of the managed memory. At the moment, it would create less problems to leave it separated with an ifdef.

PR #248, opened 2021-02-10
Reduce scope of some variables and address some cppcheck warnings

jrood-nrel:
I can't say this solves the nan issue with the builtin explicit reactions integrator, but it does seem to avoid it. While I was working on the code, I added some fixes for some warnings to this as well.

### DIFF
--- a/SourceCpp/React.cpp
+++ b/SourceCpp/React.cpp
@@ -297,9 +297,10 @@ PeleC::react_state(
               rY_in.begin());
           } else {
 #ifdef AMREX_USE_GPU
+            const int reactor_type = 1;
             react(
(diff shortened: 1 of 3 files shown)


SourceCpp/React.cpp
```diff
@@ -300,8 +298,8 @@ PeleC::react_state(
           } else {
 #ifdef AMREX_USE_GPU
             react(
-              bx, rhoY, frcExt, T, rhoE, frcEExt, fc, mask, dt, current_time,
-              reactor_type, amrex::Gpu::gpuStream());
```
hsitaram, line 304: just a minor comment, i would like to keep the reactor_type variable, just to be clear on the arguments on the call.

hsitaram (approved):

PR #249, opened 2021-02-17
PeleC hooks to SRK EOS

baperry2:
Draft for now because I anticipate that this may need some tweaks.

Includes the following:

- #ifdef code blocks around all EOS calls where real gas EOS requires additional inputs relative to ideal gas
- Additional code for differences in the diffusion calculation 
- Modified PeleC explicit reactions integrator to eliminate ideal gas assumption when using nonideal EOS (basically what was in this cancelled PR: https://github.com/AMReX-Combustion/PeleC/pull/210)
- Not supported for nonideal EOS as of now: LES, chemistry integrators from PelePhysics
- Pairs with this PelePhysics PR for calculation of nonideal transport coefficients: https://github.com/AMReX-Combustion/PelePhysics/pull/140


jrood-nrel:
I just have two comments to start off with.

Even though we're mostly stuck with SRK as a compile time choice, I think we should avoid ifdefs and use runtime logic for this. We will have to discuss the best way to do such a thing, but it will help reduce the complexity of testing. In the future, I have a vision of the mechanisms as libraries, and problem directories as classes. I am not confident we can ever get there, but PeleC has an ifdef problem we need to make an effort to halt.

Second, let's not bother updating the PelePhysics submodule in this PR. Let's wait until PelePhysics has the necessary updates in its development branch to do so. Then we don't have to do it twice.

Other than that, these changes look fine to me, and I always appreciate the work.

baperry2:
Yeah I admit the ifdefs aren't great. Most could be removed by creating appropriate wrappers in PelePhysics for Fuego and GammaLaw (e.g., wrapping TY2E with RTY2E), but we had originally decided on this path to keep it transparent in PeleC which arguments are actually being used. And that would still leave some ifdefs where there are bigger differences (transport and reactions), but would get rid of most of them. Alternatively, could add a name variable to each EOS in PelePhysics and do a runtime conditional. Also this might be impacted by more elaborate changes being discussed for LES implementation, rendering the whole thing moot.

jrood-nrel:
Ok I updated PelePhysics with your additions. Since @marchdf and I have started actively working on generalizing the EOSes, we will remove the ifdefs in the future. I will merge this if you think it's ready. You just have to merge in the development branch here.

jrood-nrel:
This looks fine to me. We want to merge it for @marchdf 's work. Are you fine with the state it's in for us to merge it?

baperry2:
Yeah I was just waiting for the tests to complete then I was going to say to merge.

PR #255, opened 2021-03-03
Updating to templated pelephysics

marchdf:
All non-reacting tests pass and reacting tests have minimal diffs. Draft for now until the PelePhysics PR (https://github.com/AMReX-Combustion/PelePhysics/pull/141) is merged.


drummerdoc:
Can you give a couple of sentences that explain why you want to template on Eos (rather than simply linking to a single one)? It sort of implies that you want access to multiple Eos's during a single run.

marchdf:
There's a couple advantages. But the easiest to point to is that it simplifies some ifdefs and code. For example, non-ideal EOS triggers a different calculation for velocity diffusion. Instead of having an ifdef there, the call site in diffterm looks the same regardless of the EOS: 
```
SpeciesEnergyFlux<pele::physics::PhysicsType::eos_type>()(i, j, k, im, jm, km, dxinv, coef, q, flx);
```
And `SpeciesEnergyFlux` is specialized on the EOS. 

In PelePhysics, similar stuff happens in transport (e.g. Chung corrections) and we can simplify things there too. 

When we implement LES models, this is going to proliferate and so this PR positions us to do LES later.

jrood-nrel:
From my side. If we can continue abstracting these things in C++ itself, we're going to make PeleC/LM much more extensible. With the current `ifdef` requirements that are necessary to add something like SRK EOS calls. When this shows up in future things like LES, we're going to hit major roadblocks with a matrix-type explosion of these. Which we already started to see for SRK. If we can use this to tackle transport and the problem types and mechanisms, we can make Pele a single executable and choose (hopefully) _anything_ at runtime. We will also save immense amounts of time by not building an executable for every test.

drummerdoc (approved):

baperry2 (approved):

jrood-nrel (approved):

PR #257, opened 2021-03-09
Explicit copy of problem parameters to device

jrood-nrel:
Here we removed the reliance on managed memory for the problem parameter structs and do the copies to device explicitly.

I'm a little concerned about diffs, but the templating for the EOS hasn't run overnight yet, so I need to compare against that first.

### DIFF
--- a/Exec/RegTests/EB-BluffBody/prob.cpp
+++ b/Exec/RegTests/EB-BluffBody/prob.cpp
@@ -17,27 +17,28 @@ amrex_probinit(
   // Parse params
   {
     amrex::ParmParse pp("prob");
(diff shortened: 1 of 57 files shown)


jrood-nrel:
The diffs I was seeing seem to have been due to merging the templating into this. So I think it's good.

marchdf:
This looks good. Would it make sense to rename `h_prob_parm_device` to something like `h_prob_parm`?

marchdf:
hmmm maybe not...

jrood-nrel:
Unfortunately I don't think so because that is the "device safe" list and it's the one that needs a host and device copy.

marchdf (approved):

PR #260, opened 2021-03-23
Mass fraction cleaning added to ctoprim

marchdf:
@hariswaran did the the work so I will let him take it from here.

### DIFF
--- a/Source/Diffusion.cpp
+++ b/Source/Diffusion.cpp
@@ -126,7 +126,6 @@ PeleC::getMOLSrcTerm(
       const amrex::Box cbox = amrex::grow(vbox, ng - 1);
       auto const& MOLSrc = MOLSrcTerm.array(mfi);
 
-
 #ifdef PELEC_USE_EB
       amrex::Real wt = amrex::ParallelDescriptor::second();
       const auto& flag_fab = flags[mfi];
--- a/Source/Diffusion.cpp
+++ b/Source/Diffusion.cpp
@@ -125,6 +125,8 @@ PeleC::getMOLSrcTerm(
       const amrex::Box gbox = amrex::grow(vbox, ng);
       const amrex::Box cbox = amrex::grow(vbox, ng - 1);
       auto const& MOLSrc = MOLSrcTerm.array(mfi);
+      
+      const int captured_clean_massfrac = clean_massfrac;
 
 #ifdef PELEC_USE_EB
       amrex::Real wt = amrex::ParallelDescriptor::second();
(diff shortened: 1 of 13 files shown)


Source/Utilities.H
```diff
@@ -199,7 +261,8 @@ pc_ctoprim(
   amrex::Array4<const amrex::Real> const& u,
   amrex::Array4<amrex::Real> const& q,
   amrex::Array4<amrex::Real> const& qa,
-  PassMap const& pmap)
+  PassMap const& pmap,
+  const int clean_massfrac = 1)
```
jrood-nrel: Why is this set to 1 here?

Source/Diffusion.cpp
```diff
@@ -126,6 +126,8 @@ PeleC::getMOLSrcTerm(
       const amrex::Box cbox = amrex::grow(vbox, ng - 1);
       auto const& MOLSrc = MOLSrcTerm.array(mfi);
 
+      const int captured_clean_massfrac = clean_massfrac;
```
jrood-nrel: Can we move all these captures to just before the `ParallelFor` in which they are used?

Source/Utilities.H
hsitaram: being consistent with our default from pelec_queries. see https://github.com/AMReX-Combustion/PeleC/blob/ba4b807d8cf13b5a691debc49a9ca1da6533abf8/Source/Params/param_includes/pelec_defaults.H#L105

Source/Diffusion.cpp
hsitaram: Its done now except in react where it is used in multiple places

Source/Utilities.H
jrood-nrel: I think we should remove the `=1` since the default or whatever is set should be passed in right? I think it needs to be added to https://github.com/AMReX-Combustion/PeleC/blob/ba4b807d8cf13b5a691debc49a9ca1da6533abf8/Source/LES.cpp#L467 as well.

hsitaram: sounds good. I have made the change.

jrood-nrel: Awesome, thanks.

jrood-nrel (approved):
Ok I'm fine with this if you guys think it should go in.

hsitaram (approved):

Source/LES.cpp
```diff
@@ -541,10 +546,12 @@ PeleC::getDynamicSmagorinskyLESTerm(
       {
         BL_PROFILE("PeleC::ctoprim()");
         const PassMap* lpmap = d_pass_map;
+        const int captured_clean_massfrac = clean_massfrac;
         amrex::ParallelFor(
           g2box, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
             pc_ctoprim(
-              i, j, k, filtered_S_ar, filtered_Q_ar, filtered_Qaux_ar, *lpmap);
+              i, j, k, filtered_S_ar, filtered_Q_ar, filtered_Qaux_ar, *lpmap,
+              clean_massfrac);
```
jrood-nrel: This should be `captured_clean_massfrac`.

PR #275, opened 2021-04-29
Hari/addheatrelease

hsitaram:
This PR adds heat-release as an extra variable to plot. Most often it gets cumbersome to calculate offline using other methods. see below heat-release from the PMF case active at the flame interface.
Screen Shot 2021-04-29 at 1 50 41 PM

### DIFF
--- a/Source/React.cpp
+++ b/Source/React.cpp
@@ -431,8 +431,20 @@ PeleC::react_state(
           bx, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
             I_R(i, j, k, NUM_SPECIES + 1) = 0.0;
             auto eos = pele::physics::PhysicsType::eos();
+
             amrex::Real hi[NUM_SPECIES] = {0.0};
+
+#ifndef PELEC_USE_SRK
             eos.T2Hi(snew_arr(i, j, k, UTEMP), hi);
+#else
+            amrex::Real Yspec[NUM_SPECIES] = {0.0};
+            for(int nsp=0;nsp<NUM_SPECIES;nsp++)
+            {
+                Yspec[nsp]=snew_arr(i, j, k, UFS+nsp)/snew_arr(i, j, k, URHO);
+            }
+            eos.RTY2Hi(snew_arr(i,j,k,URHO), snew_arr(i,j,k,UTEMP), Yspec, hi);
+#endif
+
             for (int nsp = 0; nsp < NUM_SPECIES; nsp++) {
               I_R(i, j, k, NUM_SPECIES + 1) -= hi[nsp] * I_R(i, j, k, nsp);
             }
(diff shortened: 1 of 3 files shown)


Source/React.cpp
```diff
@@ -425,6 +425,30 @@ PeleC::react_state(
...
+
+        // update heat release
+        amrex::ParallelFor(
+          bx, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
+            I_R(i, j, k, NUM_SPECIES + 1) = 0.0;
+            auto eos = pele::physics::PhysicsType::eos();
+
+            amrex::Real hi[NUM_SPECIES] = {0.0};
+
+#ifndef PELEC_USE_SRK
```
marchdf, line 437: Can we remove the ifdef on SRK and just have it do the same  RTY2Hi ? It will ignore `R` for non-SRK

jrood-nrel (approved):
This is fine with me. I can't figure out why clang-format skips the else part of the ifdef, but I will let it slide.

hsitaram:
@shashankNREL - can you check if I did the right thing for an SRK eos?

hsitaram:
> This is fine with me. I can't figure out why clang-format skips the else part of the ifdef, but I will let it slide.

The for curly braces are finally aligned the way I like it :)

marchdf (changes requested):

Source/React.cpp
hsitaram: sure, it will ignore Y as well?

shashankNREL:
Other than the changes Marc requested to remove the SRK ifdef, I am ok with these changes.

Source/React.cpp
hsitaram: RTY2Hi doesnt exist when SRK is disabled

shashankNREL: @marchdf Correct me if I am wrong but it should ignore species because in the background it will be calling 
`void speciesEnthalpy(amrex::Real * species, amrex::Real *  tc)` which doesn't need any species mass fractions.

hsitaram:
I_R(i, j, k, NUM_SPECIES + 1) = 0.0 is set to 0 and updated from within
the species loop using the "-="

marchdf (approved):

shashankNREL (approved):

Source/React.cpp
marchdf: Yeah it should be fine... if it existed. But we don't have a dummy Fuego `RTY2Hi` that mirrors `TY2Hi`. So the right fix would be to add in PelePhysics. But we have other `USE_SRK` defs in the code that will have to be lifted some time and we can address that then.

baperry2: Yeah, if there were an RTY2Hi for Fuego it would just wrap around T2Hi which in turn just wraps CKHMS(&T, Hi). It looks like we didn't implement the RTY2Hi wrapper in the Fuego EOS because it only shows up in PeleC for the nonideal transport calculation.

jrood-nrel:
I will merge this and we will address all the `PELEC_USE_SRK` later.

PR #276, opened 2021-04-30
Clean up SUNDIALS code for reactions

jrood-nrel:
Before I get any further, we need to discuss what I'm trying to do here. So far I have done this:
```
arkode 1700 to 400 LOC: 75% less code, sacrificing 10% slower on the CPU, same performance on GPU
cvode 3972 to 1686 LOC: 60% less code, (not performance tested on the CPU yet), should be same performance on GPU
```
Most of the changes are in PelePhysics. Doing a PR there right now will be too hard to read, so I'm doing it here for now. I'm pretty sure I'm going to ruffle some feathers by removing stuff either accidentally or on purpose. One of the main changes has been removing the separate CPU files in PP and using the GPU pathway for the CPU as well. This probably reduced some capabilities. I'm probably starting out by going too far, so we need to discuss what's right and what's wrong about my decisions thus far.

Edit: I have not done anything in PeleLM to match this yet, until we talk about the general changes in PelePhysics.


jrood-nrel:
Corresponding changes for PP: https://github.com/AMReX-Combustion/PelePhysics/compare/jrood/cleanup_reactions

jrood-nrel:
Ok this is ready. We need to merge https://github.com/AMReX-Combustion/PelePhysics/pull/156 before this so I can update the submodule here and then merge. The idea is that PeleC should only be changing in regards to arkode and none of these changes are expected to disturb PeleLM yet. I'm sure I will need to fix a few things after this gets merged.

PR #287, opened 2021-05-20
Fix to ensure the BCs passed to the redistribution are correct

marchdf:


Source/Utilities.H
```diff
@@ -6,6 +6,22 @@
 #include "IndexDefines.H"
 #include "PelePhysics.H"
 
+template <class T>
+amrex::Gpu::DeviceVector<T>
+convertToDeviceVector(amrex::Vector<T> v)
+{
+  int ncomp = v.size();
+  amrex::Gpu::DeviceVector<T> v_d(ncomp);
+#ifdef AMREX_USE_GPU
```
jrood-nrel: Can we use https://github.com/AMReX-Combustion/PeleC/blob/2bbdd784d35b76ccc7b033427167f457d7f29eb3/Source/PeleC.cpp#L617-L619 instead? I know I have a few of these to address.

marchdf: for sure. let me give that a shot

jrood-nrel: Other examples https://github.com/AMReX-Combustion/PeleC/blob/2bbdd784d35b76ccc7b033427167f457d7f29eb3/Exec/RegTests/PMF/prob.cpp#L100-L105

jrood-nrel (approved):

PR #294, opened 2021-06-03
Remove useless Cs calcs

marchdf:


marchdf:
@jrood-nrel I would like the golds be reblessed before merging this. I expect this to introduce zero diffs. But I want to be sure.

jrood-nrel (approved):

PR #295, opened 2021-06-08
Update AMReX submodule and remove references to VecReduce

jrood-nrel:


Source/InitEB.cpp
```diff
@@ -242,31 +242,20 @@ PeleC::initialize_eb2_structs()
...
-            int const& r, amrex::Gpu::Handler const& h) noexcept {
-            amrex::Gpu::deviceReduceSum_full(dp, r, h);
-          });
-#else
-          [=] AMREX_GPU_DEVICE(int const& r) noexcept {
-            amrex::Gpu::deviceReduceSum_full(dp, r);
-          });
-#endif
-        const int Nall_cut_faces = ds.dataValue();
+          1);
```
jrood-nrel: @WeiqunZhang it's not obvious to me what these `1` values are. Can you clarify their meaning?

Source/Utilities.H
```diff
@@ -502,14 +502,15 @@ locate(const amrex::Real* xtable, const int n, const amrex::Real& x, int& idxlo)
...
 
-/* This was manually inlined to avoid unique_ptr somehow finding it and
-segfaulting template <typename T> AMREX_GPU_DEVICE AMREX_FORCE_INLINE void
-swap(T* x, T* y)
-{
-  T temp = *x;
-  *x = *y;
-  *y = temp;
-}*/
+// This was manually inlined to avoid unique_ptr somehow finding it and
```
marchdf: Can we just remove this now?

Source/InitEB.cpp
WeiqunZhang: There should not be a 1 here.  Reduce::Sum has an optional argument for the initial value that's difaulted to zero.  We need `1` in the next use of Reduce::Sum, but not here.

jrood-nrel: Ok. Hmm, it wouldn't compile for me without a value there. I will try again.

jrood-nrel: So when I remove it, I don't understand why it doesn't use 0 as the default and fails to compile as shown in the CI.

WeiqunZhang:
Try `Reduce::Sum<int>(....)`.

jrood-nrel:
Using `<int>` seems to work. Thanks!

marchdf (approved):
Ran the tests on eagle and everything seems to pass without a diff. So I think this is good to go! I just had one comment/thing to remove.

Source/Utilities.H
WeiqunZhang: If you need to swap like this, you can also do `amrex::Swap(*x,*y)`.

PR #297, opened 2021-06-10
Add tagging documentation

marchdf:


jrood-nrel (approved):

hsitaram:
All good.. Although I notice in the code that velx,vely and velz are tagged based on value and not on absolute value. Absolute value makes more sense to me which is orientation independent. Not critical though to be done in this PR

hsitaram (approved):

marchdf:
@hsitaram good catch, I opened #298 to fix that.

PR #300, opened 2021-06-16
Add CFL warning when using MOL hydro

marchdf:


jrood-nrel (approved):

hsitaram (approved):
all good! a not-so-critical comment is that it will be nicer to have a pelec::cfl_max in params rather than hard coding a 0.3. what is here is fine.

PR #302, opened 2021-06-30
Clean Riemann solver

marchdf:
This PR:
- removes the uses of `gamma` in the Riemann solver. It was just used to compute `cs`. This is now a call to the eos.
- It also removes `rhoe` (which was an unused variable in the RS).
- These two changes led to a bunch of EOS calls being taken out 

I expect machine precision diffs.


hsitaram (approved):
looks good to me! are there any diffs?

marchdf:
I see `1e-16` diffs on EB-C10. That's about what I would expect.

jrood-nrel (approved):

baperry2 (approved):
Looks good to me.

PR #303, opened 2021-07-01
Specify exact hydro EB wall flux

marchdf:
This PR:
- sets the hydrodynamic fluxes at the EB boundary to be `F = [0, p_nx, p_ny, p_nz, 0]` (for mass, momentum, energy). The current code would call a Riemann solver with opposing normal velocities.
- removes the use of `eb_small_vfrac` as it was only being used to merge quantities in small cells for computing MOL hydro fluxes. And even then, all our cases were setting that value to 0.0.

### DIFF
--- a/Source/Diffusion.cpp
+++ b/Source/Diffusion.cpp
@@ -395,8 +395,8 @@ PeleC::getMOLSrcTerm(
             cbox, qar, qauxar, flx, area_arr, dx, plm_iorder
 #ifdef PELEC_USE_EB
             ,
-            eb_small_vfrac, vfrac.array(mfi), flags.array(mfi),
-            d_sv_eb_bndry_geom, Ncut, d_eb_flux_thdlocal, nFlux
+            flags.array(mfi), d_sv_eb_bndry_geom, Ncut, d_eb_flux_thdlocal,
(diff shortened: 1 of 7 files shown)


marchdf:
C14: left is this PR, right is development:
Screen Shot 2021-06-30 at 4 41 57 PM

marchdf:
C7: 
[profiles.pdf](https://github.com/AMReX-Combustion/PeleC/files/6749196/profiles.pdf)

marchdf:
C8 (left is PR, (right is `development`):
Screen Shot 2021-07-01 at 10 20 39 AM

marchdf:
C9: 
[profiles.pdf](https://github.com/AMReX-Combustion/PeleC/files/6749626/profiles.pdf)

jrood-nrel (approved):

hsitaram (approved):
so no more merge-cell business then?

marchdf:
@hsitaram this was the only place we had the possibility of merging cells. But we weren't merging anymore because of the new State Redistribution scheme. And if we use this EB wall flux, where pressure is the only variable that is used, then I don't think merging pressure makes any sense...

PR #305, opened 2021-07-06
Hari/least squares

hsitaram:
This PR implements the least-squares stencil for calculating gradients at the embedded boundary for diffusion operators.
I have done some preliminary tests and it seems to be giving accurate results. see EB-MMS error convergence
errconv_ebc1


Source/InitEB.cpp
```diff
@@ -84,6 +91,8 @@ PeleC::initialize_eb2_structs()
     const amrex::Box tbox = mfi.growntilebox();
     const amrex::EBCellFlagFab& flagfab = flags[mfi];
 
+    // amrex::Print()<<"grown tbox:"<<tbox<<"\n";
```
marchdf: remove these?

marchdf:
Super cool! Can you add some documentation? Maybe include the figure you posted above in the documentation too?

hsitaram:
> Super cool! Can you add some documentation? Maybe include the figure you posted above in the documentation too?

I can definitely add to documentation in the next PR, now that I see ~ 2nd order accuracy with EB which is big! Doing a 512 case as well.

Source/InitEB.cpp
```diff
@@ -76,6 +77,14 @@ PeleC::initialize_eb2_structs()
   int bgs = -1;
   pp.get("boundary_grad_stencil_type", bgs);
 
+  if (bgs == 0) {
```
jrood-nrel, line 79: I think this should be an if/else if.

Source/EBLeastSquares.H
```diff
@@ -0,0 +1,296 @@
...
+// data on boundaries of control volumes, the unweighted least squares procedure
+// is used. When actual gradients are required, as in the production terms for
+// the turbulence models, Green’s theorem"
+
+//"I guess the upshot from Anderson and Bonhaus's statement is that
+// least-squares is robust and more accurate on stretched weird grids but may
+// not
+// be necassarily more accurate than Green-Gauss on nicer grids": HARI S
+
+AMREX_GPU_DEVICE
```
jrood-nrel: I think these should be `AMREX_FORCE_INLINE` if they are device functions?

Source/InitEB.cpp
```diff
@@ -1,6 +1,7 @@
 #include <memory>
 
 #include "EB.H"
+#include "EBLeastSquares.H"
```
jrood-nrel: I don't think this warrants a new header. Can this go in the existing EB files?

Source/InitEB.cpp
```diff
@@ -76,6 +77,14 @@ PeleC::initialize_eb2_structs()
   int bgs = -1;
   pp.get("boundary_grad_stencil_type", bgs);
 
+  if (bgs == 0) {
+    amrex::Print() << "using quadratic stencil\n";
```
jrood-nrel: Capitalize `using`

jrood-nrel:
I just had a few nitpicks and then I'm fine merging this.

Source/InitEB.cpp
hsitaram: I have changed this

Source/EBLeastSquares.H
hsitaram: good catch! added now

Source/InitEB.cpp
hsitaram: done

Source/InitEB.cpp
hsitaram: I could split the least-squares specific helper functions into Utilities.H or other headers, but I feel it will be hard to read. If this stays as a stand-alone header a developer has to just go through this one file. It will also not blow up the other files. So I prefer that it remain in a separate header.

Source/InitEB.cpp
hsitaram: I also made it a query, so it doesnt err if not specified. dont know why it remained as pp.get

Source/InitEB.cpp
jrood-nrel: There are already a lot of EB utility functions in `EB.H` and `EB.cpp` so I think more would be fine. I feel there is an inconsistency with `pc_fill_bndry_grad_stencil_ls` being in `EBLeastSquares.H` and `pc_fill_bndry_grad_stencil_quadratic` in `EB.cpp` when they are so similar. Why not an `EBQuadratic.H` as well? (not that I'm suggesting that 🙂 )

hsitaram (commented):
I have added the least-squares functions into EB.H and EB.cpp. No additional headers.

marchdf:
@hsitaram can we still add the documentation? You mentioned you wanted to add that. I think the graph you show here should be in that. Maybe in the EB discussion?

hsitaram:
> @hsitaram can we still add the documentation? You mentioned you wanted to add that. I think the graph you show here should be in that. Maybe in the EB discussion?

yes, as soon as my 512x512x512 case is done, I will add it.

PR #306, opened 2021-07-08
Add converging nozzle flow

marchdf:


jrood-nrel (approved):

Exec/RegTests/EB-ConvergingNozzle/prob.H
```diff
@@ -0,0 +1,126 @@
...
+  state(i, j, k, UTEMP) = T;
+  for (int n = 0; n < NUM_SPECIES; n++) {
+    state(i, j, k, UFS + n) = rho * massfrac[n];
+  }
+}
+
+AMREX_GPU_DEVICE
+AMREX_FORCE_INLINE
+void
+bcnormal(
```
baperry2, line 64: I have a version of this that has the subsonic outlet BC from Blazek implemented here as well (based on EB-C10). With the current FOExtrap BC, the pressure in the nozzle builds over time (it is not defined at either the inlet or outlet), but the Blazek BC holds it constant. However, in the current BC seems to be better at letting pressure waves escape the domain in test runs that I did. The outlet BC is not really relevant for what this test is targetting so it may not matter much how we specify the outlet BC.

Exec/RegTests/EB-ConvergingNozzle/README.md
```diff
@@ -0,0 +1 @@
+# Simple flow in a converging nozzle geometry
```
baperry2: Maybe specify subsonic here + how BCs are defined.

Mention how to check for the problem that led to this test being added?

baperry2 (approved):
Looks good for now, may be worth revisiting the outlet BC in this and other cases at some point.

Exec/RegTests/EB-ConvergingNozzle/prob.H
drummerdoc: This is a good point, and might be responsible for weird looking flows I'm seeing in a different configuration.  @emotheau has suggested to fix the pressure at the outflow, extrap T, U and composition, as a simple approach to what it seems like @baperry2 is suggesting.

marchdf:
I think this is ready to merge. AMReX-Hydro has fixes that address the issue.  There was apparently a bug in `StateRedist` that resolved 98% of the issues. And then there was the addition of `NewStateRedist` which is even better. Eventually `NewStateRedist` will become `StateRedist` in the future. Here's a comparison of the different schemes/commits:

```
top left: latest AMReX-Hydro with “NewStateRedist”
top right: latest AMReX-Hydro with “StateRedist”
bottom left: f986086 for AMReX-Hydro with “StateRedist”
bottom right: 86561b1 for AMReX-Hydro with “StateRedist”
```
Screen Shot 2021-08-12 at 9 21 00 AM

 As can be seen, the oldest `StateRedist` (bottom right) has a cross like pattern in velocity. Newer `StateRedist` and `NewStateRedist` don't have this issue.

drummerdoc:
I'd say that was an improvement!

marchdf:
For sure!! I will be trying this out in the pistonbowl later...

drummerdoc:
Oops, finger slipped on the re-review button... :)

drummerdoc (approved):

PR #320, opened 2021-07-26
Adds user defined inputs to set_problem_tags

paroomk:
These changes make it possible to have user defined inputs (example: a Tensorflow model) for tagging.

### DIFF
--- a/Source/Tagging.H
+++ b/Source/Tagging.H
@@ -3,6 +3,7 @@
 
 #include <AMReX_FArrayBox.H>
 #include <AMReX_TagBox.H>
+#include "prob_parm.H"
 
 struct TaggingParm
 {
@@ -140,7 +141,8 @@ struct EmptyProbTagStruct
     const amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> /*dx*/,
     const amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> /*prob_lo*/,
     const amrex::Real /*time*/,
-    const int /*level*/) noexcept
+    const int /*level*/,
+    ProbParmDevice const& d_prob_parm_device) noexcept
   {
   }
 };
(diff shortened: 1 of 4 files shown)


Source/Tagging.H
```diff
@@ -140,7 +141,8 @@ struct EmptyProbTagStruct
     const amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> /*dx*/,
     const amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> /*prob_lo*/,
     const amrex::Real /*time*/,
-    const int /*level*/) noexcept
+    const int /*level*/,
+    ProbParmDevice const& d_prob_parm_device) noexcept
```
marchdf: can you just add `/* ... */` around `d_prob_parm_device`? Here and in the two other prob.H files edited.

marchdf (changes requested):

Source/Tagging.H
drummerdoc: @marchdf et al, what do you think about adding the generic tagging routines into PeleC that allow configuring a bunch of them in the inputs at runtime (rather than compile time), as done in PeleLM?  (search for ``amr.refinement_indicators`` in the Exec/RegTests folder for example usage)

marchdf: would be very neat to add that too!

baperry2: I do think it would be ideal to at least have the same syntax for setting runtime refinement criteria between the two codes. PeleLM's is definitely more elegant and flexible. But the existing PeleC approach does seem to enable all types of refinement that actually get used, so it seems fine for now.

In any event, the target of this pull request is somewhat separate from that discussion.

marchdf: Absolutely.

drummerdoc: @baperry2 Yes, it is separate.

baperry2 (approved):
Looks good, thanks Parvathi.

marchdf (approved):

PR #321, opened 2021-07-26
No chem source in E eqn in docs

baperry2:
The chemical reaction term in the energy equation in the documentation seems to be extraneous, so this PR removes it. 

On a related note, we compute an entry in I_R for total energy: https://github.com/AMReX-Combustion/PeleC/blob/8f248555c1e7d809597049a18d024c4045c5d1c9/Source/React.cpp#L288, but as far as I can tell it reduces to 0 (within numerical precision), as I would expect, at least for a purely gas phase system. Is there any reason not to simplify the code and simply set the value to 0 directly?

### DIFF
--- a/Docs/sphinx/Equations.rst
+++ b/Docs/sphinx/Equations.rst
@@ -24,7 +24,7 @@ PeleC advances the following set of fully compressible equations for the conserv
 
   \frac{\partial (\rho \mathbf{u})}{\partial t} &=& - \nabla \cdot (\rho \mathbf{u} \mathbf{u} -p{\mathcal I} + \Pi) - \nabla p +\rho \mathbf{g} + \mathbf{S}_{{\rm ext},\rho\mathbf{u}},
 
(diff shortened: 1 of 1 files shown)


drummerdoc:
Right, it's zero by virtue of the definition of the internal energy, which for combustion applications following the chemkin standard, includes the molecular fiducial energy. I don't think it harms anything to keep those terms there and it makes extension to other thermo standards maybe somewhat less error prone.  But I don't have a strong opinion either way.

baperry2:
@drummerdoc Ok, so no changes to the code then. I added a bit more discussion to the documentation similar to what is in the PeleLM docs which should hopefully avoid any confusion about the energy definition.

drummerdoc:
thanks for doing that.  It's always good to include some text on this stuff to clarify things.

hsitaram (approved):

jrood-nrel (approved):

PR #322, opened 2021-08-03
Simplify EOS and Transport structure

marchdf:


Source/React.cpp
```diff
@@ -317,7 +317,7 @@ PeleC::react_state(
 
             amrex::Real hi[NUM_SPECIES] = {0.0};
 
-#ifndef PELEC_USE_SRK
```
baperry2, line 320: While you're updating EOS stuff, how about knocking out this ifdef by adding an eos.RTY2Hi wrapper around eos.T2Hi for Fuego (as for other eos functions) and just calling eos.RTY2Hi no matter what?

Source/Particle.cpp
```diff
@@ -191,7 +191,7 @@ PeleC::defineParticles()
   if (SPRAY_FUEL_NUM > NUM_SPECIES) {
     Abort("Cannot have more spray fuel species than fluid species");
   }
-#ifdef PELEC_EOS_FUEGO
```
baperry2, line 194: Not familiar with this part of the code or if coupling sprays and SRK is meant to be supported at all, but if so it seems to me like SRK should be treated the same as Fuego rather than the same as GammaLaw here.

Source/React.cpp
marchdf: Excellent idea. I implemented that.

Source/Particle.cpp
marchdf: good catch!

Exec/Make.PeleC
```diff
@@ -46,9 +46,15 @@ Bpack             += $(PP_SRC_HOME)/Make.package
 Blocs             += $(PP_SRC_HOME)
 
 # EOS
-EOS_HOME = $(PELE_PHYSICS_HOME)/Eos/$(strip $(Eos_dir))
-ifeq ($(Eos_dir),$(filter $(Eos_dir),Soave-Redlich-Kwong))
-   DEFINES += -DPELEC_USE_SRK
+EOS_HOME = $(PELE_PHYSICS_HOME)/Eos/
```
jrood-nrel, line 49: Will we be able to just include the `Make.PelePhysics` here?

marchdf: Yes we should consolidate those... does it make sense in this PR? There's probably a bunch more from the file that should be consolidated.

jrood-nrel: It can be done later if it's not trivial.

PR #328, opened 2021-08-17
Rk64fixes in pelec

hsitaram:
updating submodules and bringing in pelephysics' rk64 integrator into pelec. I have a feeling cmake is going to complain. lets see how CI goes.


jrood-nrel:
I think `React.H` is not needed anymore?

hsitaram:
> I think `React.H` is not needed anymore?

We will need to phase out at some point. Lets keep it until @baperry2 's SRK mods go into pelephysics

jrood-nrel:
@baperry2 Can you put the SRK modifications into PelePhysics so we don't have two rk64 solvers?

baperry2:
@jrood-nrel @hsitaram sure, the modifications should be pretty quick, I'll do them later today or tomorrow.

hsitaram:
> @jrood-nrel @hsitaram sure, the modifications should be pretty quick, I'll do them later today or tomorrow.

It is a bit tricky now that there is only one fkernelspec. are you planning to do with ifdef decorations?

baperry2:
@hsitaram having only one fKernelSpec actually makes it easier (see this [comment](https://github.com/AMReX-Combustion/PelePhysics/pull/187#issuecomment-899765339)) and should enable the changes to make sundials SRK compatible as well. It will all be handled by EOS templating and won't require any ifdefs. That may incur a slight performance penalty for SRK vs. rearranging to minimize EOS calls, but I think the simplicity is probably worth it at this point.

hsitaram:
@baperry2 - you would need an ifdef for using energy equation ode instead of temperature, correct? otherwise, cvode's analytic jacobian will not be compatible like you said in your earlier comment. You also will have to change how the sundials nvecs are initialized if e/h is the dependent variable instead of T

baperry2:
My plan is just to leave the integration based on temperature. I do think the stability benefit of including temperature as a dependent variable for the ODE solver is outweighed by the computational cost of the additional EOS calls required to do so (for SRK but not ideal gas), but it wouldn't hurt beyond that. 

I might play around to verify the relative computational cost a bit more. If it's a large performance penalty, I can use runtime conditionals to skip computation of the RHS for T for SRK (should not require any changes to nvecs structure because we manually integrate e and T would just be ignored). But I think that's a secondary consideration at this point. 

There's no short term plan to support analytical jacobian for SRK regardless - it will only be compatible with sundials for schemes that don't rely on the analytical jacobian.

drummerdoc:
@baperry2 I've also heard that working with the composition in moles/vol can make a difference as well, since we would avoid converting from rho.Y to evaluate the production rates (though it probably depends on the details under the covers).  However, the potential savings of either optimization might be minimal on GPUs (because flops are "free" as they say) - or the extra registers to hold the converted quantities could reduce occupancy... GPUs are fun.  It's certainly worth a trade-off study someday down the line, for both IGL and SRK.

CMake/BuildPeleCExe.cmake
```diff
@@ -91,6 +91,19 @@ function(build_pelec_exe pelec_exe_name)
     target_sources(${pelec_exe_name} PRIVATE
                    ${SRC_DIR}/React.H
                    ${SRC_DIR}/React.cpp)
+    if("${PELEC_SUNDIALS_INTEGRATOR}" STREQUAL "rk64")
```
jrood-nrel: We can simplify this. I will make a PR to this branch that updates the CMake.

jrood-nrel (approved):

PR #329, opened 2021-08-18
Simplify CMake

jrood-nrel:


hsitaram (approved):
Thank you Jon! This is great. I was hoping to do something like this but ended doing the bare minimum, as I am not a cmake expert.

PR #334, opened 2021-08-26
removing pelec_d_term_reverse

hsitaram:
using amrex_d_term instead for all stencils for better readability and memory access.

### DIFF
--- a/Source/EB.cpp
+++ b/Source/EB.cpp
@@ -432,9 +432,9 @@ pc_fill_flux_interp_stencil(
       const amrex::Real act0 = amrex::Math::abs(ct0);
       const amrex::Real act1 = amrex::Math::abs(ct1);
       sten[L].val[1][1] = fa(iv) * (1.0 - act0) * (1.0 - act1);
-      sten[L].val[1][t0n + 1] = fa(iv) * act0 * (1.0 - act1);
-      sten[L].val[t1n + 1][1] = fa(iv) * (1.0 - act0) * act1;
-      sten[L].val[t1n + 1][t0n + 1] = fa(iv) * act0 * act1;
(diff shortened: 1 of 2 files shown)


marchdf:
This is great! I don't know why clang-tidy is complaining @jrood-nrel any ideas? It really feels like a false positive? 

I tried EB-C1 and C14 and get no diffs. 

I do see that C14 `nan`s for both this branch and `development` after running hundreds of steps. But I think that's another issue...

marchdf (approved):

marchdf:
I am addressing the `nan` issue in PR #336

jrood-nrel:
It wants a modern `for` loop on line 382 of `EB.cpp`. I neither agree nor disagree with it. You can just put `// NOLINT` at the end of that line.

marchdf:
Yeah but why not on the next line too then? And what would that modern for loop look like?

jrood-nrel:
`NOLINT` is on the line below.

jrood-nrel:
I think it would be something like `for (int kk : grad_stencil[L].val)`, but I don't know how to do the multidimensional range based for loop in c++11.

marchdf:
It's on the `kk` line, not the `jj` line.

hsitaram:
ok nolinted it. lets see what happens

marchdf:
Oh now I want to do:
```
for( auto &r0 : grad_stencil[L].val)
    for(auto &r1 : r0)
        for(auto &r2 : r1)
          r2 = 0.0; 
```

jrood-nrel:
Hah, I think it could actually remove the need for `D_TERM` huh?

marchdf:
huh good point... maybe it would!

jrood-nrel (approved):

PR #335, opened 2021-08-26
Local derive container has components only for results, not full state

drummerdoc:


marchdf:
I am totally confused. I fixed this in PR #306 .

marchdf:
What am I missing? `development` branch shows it to be fixed: https://github.com/AMReX-Combustion/PeleC/blob/development/Source/Derive.cpp#L419

But this PR is not showing a conflict with `development`. So how is this PR possible?

drummerdoc:
I must have accidentally also included the fix in the turn_inflow merge.  Sorry about that.

PR #336, opened 2021-08-27
Update PelePhysics, AMReX and AMReX-Hydro submodules

marchdf:
Both submodules have necessary fixes for EB. For example, EB-C14 will `nan` without these changes.


jrood-nrel (approved):

marchdf:
@jrood-nrel I don't know how to remove `readability-static-accessed-through-instance`. So I added `NOLINT`. I tried the suggested fixes when googling but nothing helped. Happy to take suggestions.

marchdf:
I still need to track down the FPEs...

jrood-nrel:
That's fine. We don't need to try too hard to satisfy clang-tidy. Sometimes it's not possible.

PR #339, opened 2021-09-07
Adding Lax Friedrich's flux

hsitaram:
It is only slightly more diffusive than MOL. This scheme could be more robust with complicated EOSs. See below for SOD shock tube density solution with a close up at the contact discontinuty.

compare
fig1

### DIFF
--- a/Source/Diffusion.cpp
+++ b/Source/Diffusion.cpp
@@ -391,7 +391,7 @@ PeleC::getMOLSrcTerm(
 #endif
           // auto const& vol = volume.array(mfi);
           pc_compute_hyp_mol_flux(
-            cbox, qar, qauxar, flx, area_arr, dx, plm_iorder
+            cbox, qar, qauxar, flx, area_arr, dx, plm_iorder,use_laxf_flux
 #ifdef PELEC_USE_EB
             ,
             flags.array(mfi), d_sv_eb_bndry_geom, Ncut, d_eb_flux_thdlocal,
--- a/Source/Diffusion.cpp
+++ b/Source/Diffusion.cpp
@@ -391,7 +391,7 @@ PeleC::getMOLSrcTerm(
 #endif
           // auto const& vol = volume.array(mfi);
           pc_compute_hyp_mol_flux(
-            cbox, qar, qauxar, flx, area_arr, dx, plm_iorder,use_laxf_flux
+            cbox, qar, qauxar, flx, area_arr, dx, plm_iorder, use_laxf_flux
 #ifdef PELEC_USE_EB
             ,
             flags.array(mfi), d_sv_eb_bndry_geom, Ncut, d_eb_flux_thdlocal,
--- a/Source/MOL.cpp
+++ b/Source/MOL.cpp
@@ -153,8 +153,7 @@ pc_compute_hyp_mol_flux(
             qtempl[R_P], spl, qtempr[R_RHO], qtempr[R_UN], qtempr[R_UT1],
             qtempr[R_UT2], qtempr[R_P], spr, bc_test_val, cavg, ustar,
             flux_tmp[URHO], flux_tmp[f_idx[0]], flux_tmp[f_idx[1]],
-            flux_tmp[f_idx[2]], flux_tmp[UEDEN], flux_tmp[UEINT], tmp0, tmp1,
-            tmp2, tmp3, tmp4);
+            flux_tmp[f_idx[2]], flux_tmp[UEDEN], flux_tmp[UEINT]);
         }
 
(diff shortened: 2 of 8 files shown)


drummerdoc:
In this context, is "SDC" actually second-order Godunov with plm slopes?

hsitaram:
Sdc is with parabolic reconstruction

Source/Diffusion.cpp
```diff
@@ -391,7 +391,7 @@ PeleC::getMOLSrcTerm(
 #endif
           // auto const& vol = volume.array(mfi);
           pc_compute_hyp_mol_flux(
-            cbox, qar, qauxar, flx, area_arr, dx, plm_iorder
+            cbox, qar, qauxar, flx, area_arr, dx, plm_iorder, use_laxf_flux
```
marchdf, line 394: instead of adding a bool for `use_laxf_flux`, can we reuse/update the `riemann_solver` param?

Source/MOL.cpp
```diff
@@ -137,13 +138,24 @@ pc_compute_hyp_mol_flux(
...
         amrex::Real tmp4 = 0.0;
-        riemann(
-          qtempl[R_RHO], qtempl[R_UN], qtempl[R_UT1], qtempl[R_UT2],
-          qtempl[R_P], spl, qtempr[R_RHO], qtempr[R_UN], qtempr[R_UT1],
-          qtempr[R_UT2], qtempr[R_P], spr, bc_test_val, cavg, ustar,
-          flux_tmp[URHO], flux_tmp[f_idx[0]], flux_tmp[f_idx[1]],
-          flux_tmp[f_idx[2]], flux_tmp[UEDEN], flux_tmp[UEINT], tmp0, tmp1,
-          tmp2, tmp3, tmp4);
+
+        if (!use_laxf_flux) {
```
marchdf, line 142: There should be an equivalent riemann call in the pc_umdrv stuff for PPM/PLM. Can we update there too?

Source/Riemann.H
```diff
@@ -171,4 +171,94 @@ riemann(
...
+  const int,
+  const amrex::Real,
+  amrex::Real& ustar,
+  amrex::Real& uflx_rho,
+  amrex::Real& uflx_u,
+  amrex::Real& uflx_v,
+  amrex::Real& uflx_w,
+  amrex::Real& uflx_eden,
+  amrex::Real& uflx_eint,
+  amrex::Real&,
```
baperry2: Why not just delete these unused arguments all together?

Source/Params/param_includes/pelec_defaults.H
```diff
@@ -35,6 +35,7 @@ int PeleC::ppm_reference_eigenvectors = 0;
 int PeleC::plm_iorder = 2;
 int PeleC::hybrid_riemann = 0;
 int PeleC::use_colglaz = -1;
+int PeleC::use_laxf_flux = 0;
 int PeleC::riemann_solver = 0;
 int PeleC::cg_maxiter = 12;
```
baperry2, line 40: Unrelated but it seems like there's a bunch of stuff in here for deprecated Riemann solvers that might be able to be cleaned up

baperry2 (approved):

marchdf (changes requested):

Source/MOL.cpp
hsitaram: This new scheme is only to determine the flux and hence cannot be used as a riemann solver directly into the time-centered godunov discretization which has those "notorious" transverse terms.  This scheme can be used within MOL update as only a flux is required without interface Godunov states.

Source/Diffusion.cpp
hsitaram: Strictly speaking laxfriedrich_flux is not a riemann solver. Right now it is set up only to be used within MOL update. I am inclined to keep it as a separate flag, but if you insist I can change it.

Source/Params/param_includes/pelec_defaults.H
hsitaram: I think we will add it in one day :)

Source/Riemann.H
hsitaram: actually I agree. I can do this if @marchdf agrees that this is not a Riemann solver. I was just trying to be consistent with the riemann function call.

marchdf: fine by me!

Source/Diffusion.cpp
marchdf: ah gotcha. Yeah that's fine.

Source/Riemann.H
hsitaram: made this change now!

marchdf (approved):

PR #343, opened 2021-09-14
Tag all cut cells.

marchdf:
The previous version of the code has issues for coordinate aligned EB geometry. 

For example: 
This box with coordinate aligned geometry does not get refined on the finer level. Because the current tagging is on 0<vfrac<1, the coarse level is never tagged for refinement. 
Screen Shot 2021-09-14 at 10 28 01 AM

However, this PR introduces an additional loop on cut cells and those are tagged for refinement. The same case now leads to this refinement:
Screen Shot 2021-09-14 at 10 30 21 AM

### DIFF
--- a/Source/PeleC.cpp
+++ b/Source/PeleC.cpp
@@ -1758,6 +1758,16 @@ PeleC::errorEst(
           tilebox, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
             tag_error_bounds(i, j, k, tag_arr, vfrac_arr, 0.0, 1.0, tagval);
           });
+
+        const int local_i = mfi.LocalIndex();
+        const int Nebg = (!eb_in_domain) ? 0 : sv_eb_bndry_geom[local_i].size();
+        EBBndryGeom* ebg = sv_eb_bndry_geom[local_i].data();
+        amrex::ParallelFor(Nebg, [=] AMREX_GPU_DEVICE(int L) {
+          const auto& iv = ebg[L].iv;
+          if (tilebox.contains(iv)) {
+            tag_arr(iv) = tagval;
+          }
+        });
       }
 #endif
 


Source/PeleC.cpp
```diff
@@ -1758,6 +1758,18 @@ PeleC::errorEst(
...
+        const int Nebg = (!eb_in_domain) ? 0 : sv_eb_bndry_geom[local_i].size();
+        EBBndryGeom* ebg = sv_eb_bndry_geom[local_i].data();
+        const auto lo = amrex::lbound(tilebox);
+        const auto hi = amrex::ubound(tilebox);
+        amrex::ParallelFor(Nebg, [=] AMREX_GPU_DEVICE(int L) {
+          const auto& iv = ebg[L].iv;
+          if (is_inside(iv, lo, hi)) {
+            tag_arr(iv) = tagval;
+          }
+        });
```
drummerdoc, line 1770: Why can't we just run through the list of ebbndrygeom objects and tag if tilebox.contains(ebg[L].iv)? Not sure  understand why we have to do all that other checking.

marchdf: It's what happens when you do this between meetings ;) I updated the code. Is that what you meant?

drummerdoc: Yeah, exactly!

hsitaram (approved):
just curious. does ebg pick up those edge cells as something thats not regular and not covered in this specific case where the EB aligns with the grid? It seems to do so I think. Safest way could be tag based on vfrac_grad.

marchdf:
If you are referring to the cells I think you are, ebg calls those cells cut because they have a face that has an afrac of 0.

drummerdoc:
I think this looks good now, but I'd like the Norway folks to give it a test (and they're offline at the moment).

hsitaram:
@drummerdoc - merge this in?

drummerdoc:
In the SINTEF problem, this currently gives weird results, so I'd like to sort out whether it is an issue with this or with the SINTEF geometry setup.

drummerdoc:
This looks good with tests from the SINTEF work.

drummerdoc (approved):

PR #344, opened 2021-09-15
Fix derived variables with gradients at EB's

marchdf:
Quantities derived from gradients were incorrect at EB boundaries. They would use the state body values and we would see weirdness. This affected `divu`, `enstrophy`, and `magvort`. This PR fixes that.

A good illustration is the EB-C14 case (shock-ramp interaction).

Before this PR:
Screen Shot 2021-09-15 at 3 54 06 PM

With this PR:

Screen Shot 2021-09-15 at 3 52 25 PM

### DIFF
--- a/Source/Derive.H
+++ b/Source/Derive.H
@@ -8,6 +8,56 @@
 using namespace MASA;
 #endif
 
+#ifdef PELEC_USE_EB
+#include "EB.H"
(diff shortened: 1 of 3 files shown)


hsitaram (approved):
nice! I like the one-sided difference idea

PR #347, opened 2021-09-21
Enable restarting from inert run and other fixes

### LINKED ISSUE #304: Restarting a reacting simulation from an inert checkpoint file
If a checkpoint file has been generated by the code compiled with `USE_REACT = FALSE`, the code fails to restart from this checkpoint file if the code is recompiled with `USE_REACT = TRUE`.

The output throw this error:

```
restarting calculation from file: chk0000386000
Starting to call amrex_probinit ...
13 variables found in PMF file
2 data lines found in PMF file
Successfully run amrex_probinit
Successfully read inputs file ...
amrex::Error::5::operator>>(istream&,Box&): expected '(' !!!
SIGABRT
```

I suspect that when the code is compiled with `USE_REACT = TRUE`, the code is looking for Reactions_Type data that are missing in the checkpoint file. I tried to add a `state_in_checkpoint[Reactions_Type] = 0` in the routine `PeleC::set_state_in_checkpoint` in `IO.cpp` to force the reader to ignore missing data, but it is unsuccessful.

marchdf:
- Enabling restarting with `do_react=1` from a simulation with `do_react=0` (and vice versa)
- Removing Reactions_Type from plt and chk files when `do_react=0`
- Remove `ebmask` in favor of directly using `flag_fab` (it's been a TODO since forever)
- Fix the FAB ops (particularly the setval) to operate on the box we want and not on the FAB (this is important because tiling)

Closes #304

### DIFF
--- a/Source/Derive.cpp
+++ b/Source/Derive.cpp
@@ -247,7 +247,7 @@ pc_dermagvort(
   const auto& flag_fab = amrex::getEBCellFlagFab(derfab);
   const auto& typ = flag_fab.getType(bx);
   if (typ == amrex::FabType::covered) {
-    derfab.setVal<amrex::RunOn::Device>(0.0);
+    derfab.setVal<amrex::RunOn::Device>(0.0, bx);
     return;
   }
   const auto& flags = flag_fab.const_array();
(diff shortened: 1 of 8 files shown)


drummerdoc (approved):
So, in this version, like the last, there is no way to read reaction data ON restart still, but that state data component is always created (regardless of do_react), correct?

drummerdoc (approved):
Ok, I get it.  Looks good.  Will test.

PR #349, opened 2021-09-29
Enabling better user defined EB geometry

marchdf:
Uses similar ideas as we do for runtiming the reactor selection. This runtimes the EB selection for user defined geometries. An example of how it is done is in `EB-ConvergingNozzle`.


jrood-nrel (approved):
Makes total sense.

drummerdoc (approved):
Same great taste, half the calories!

PR #350, opened 2021-09-29
Fix for cdash sanitizer

marchdf:


marchdf:
close in favor of #351

PR #370, opened 2021-11-02
Update PMF reading to use PelePhysics

marchdf:


jrood-nrel:
This doesn't run for me without managed memory on and it seems to be a few times slower.

marchdf:
well bummer.

esclapez:
I don't have a comparison of how slow in LM since it was already using this machinery, but I'll test the no managed mem. now.

esclapez:
Provided a couple of fixes in IAMR/amrex to remove RunOn::Host, LM FlameSheet using PelePhysics PMF runs fine on Eagle with `amrex.the_arena_is_managed = 0` (but I get a `pure virtual method called` during the finalize that I need to investigate).

PR #372, opened 2021-11-04
Add dodecane_lu PMF

jrood-nrel:


Exec/RegTests/CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ if(PELEC_ENABLE_AMREX_EB)
   add_subdirectory(EB-C10)
   add_subdirectory(EB-C11)
   add_subdirectory(EB-C12)
-  add_subdirectory(EB-C14)
+  #add_subdirectory(EB-C14)
```
marchdf, line 29: Is this intended?

jrood-nrel: Yeah EB-C14 test is disabled due to FPEs so I figure why bother building it.

marchdf: oh right

jrood-nrel:
I decided to remove the copies of data files, which means I had to break the test input files outside of cmake by adding in paths generated at configure time.

Exec/RegTests/HIT/tests/hit-1/hit-1.i
```diff
@@ -57,7 +57,7 @@ amr.plot_vars  =  density Temp
 amr.derive_plot_vars = x_velocity y_velocity z_velocity magvel magvort pressure
 
 # PROBLEM PARAMETERS
-prob.iname = "hit-1.ic"
+prob.iname = "@CMAKE_SOURCE_DIR@/Exec/RegTests/HIT/hit_ic_4_32.dat"
```
marchdf: I get that this makes is to we don't have to copy files. But it also means that the only way to run the case is through `ctest` right? Bringing this up because sometimes I run these cases outside of `ctest`...

jrood-nrel: You have to do `prob.iname = "hit_ic_4_32.dat"` on the command line, which is annoying, but the 1MB dodecane PMF file made me realize we shouldn't keep copies around. I tried symlinks but they require the developer to list the files for each test that need to be symlinked. I guess the best option might be to put the test input files back into the top level executable directory, but that poses some other issues. We should probably talk about the best way to do this next week.

marchdf: yeah sounds like there's no perfect solution. let's talk

marchdf (approved):

PR #374, opened 2021-11-09
Adding typical values functionality

hsitaram:
This PRs allows pelec to set typical values for chemistry integration with cvode/arkode

### DIFF
--- a/Source/PeleC.H
+++ b/Source/PeleC.H
@@ -697,7 +697,6 @@ protected:
   static bool use_typical_vals_chem_usr;
   static amrex::Real typical_rhoY_val_min;
   static int reset_typical_vals_int;
-  amrex::Vector<amrex::Real> typical_values_chem;
   static amrex::Vector<amrex::Real> typical_values_chem_usr;
 
   // problem-specific includes
--- a/Source/PeleC.H
+++ b/Source/PeleC.H
@@ -693,8 +693,8 @@ protected:
 
   static amrex::Vector<int> src_list;
 
-  static bool use_typ_vals_chem;
-  static bool use_typ_vals_chem_usr;
+  static bool use_typical_vals_chem;
+  static bool use_typical_vals_chem_usr;
   static amrex::Real typical_rhoY_val_min;
   static int reset_typical_vals_int;
   amrex::Vector<amrex::Real> typical_values_chem;
--- a/Source/PeleC.H
+++ b/Source/PeleC.H
@@ -230,6 +230,8 @@ public:
   void zero_in_body(amrex::MultiFab& S) const;
 #endif
 
+  void set_typical_values_chem();
+
   // Proceed with next timestep?
   virtual int okToContinue() override;
 
@@ -690,6 +692,13 @@ protected:
   static amrex::Vector<std::string> spec_names;
 
   static amrex::Vector<int> src_list;
+
+  static bool use_typ_vals_chem;
+  static bool use_typ_vals_chem_usr;
+  static amrex::Real typical_rhoY_val_min;
+  static int reset_typical_vals_int;
+  amrex::Vector<amrex::Real> typical_values_chem;
+  static amrex::Vector<amrex::Real> typical_values_chem_usr;
 
   // problem-specific includes
 #include <Problem.H>
--- a/Source/PeleC.cpp
+++ b/Source/PeleC.cpp
@@ -109,7 +109,7 @@ int PeleC::num_state_type = 0;
 amrex::Real PeleC::typical_rhoY_val_min = 1.e-10;
 bool PeleC::use_typical_vals_chem = false;
 bool PeleC::use_typical_vals_chem_usr = false;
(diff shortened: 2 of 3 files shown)


Source/PeleC.H
```diff
@@ -691,6 +693,13 @@ protected:
 
   static amrex::Vector<int> src_list;
 
+  static bool use_typ_vals_chem;
```
marchdf: for consistency with the other stuff you wrote: `typ` -> `typical` (here and below)

Source/PeleC.cpp
```diff
@@ -348,6 +366,8 @@ PeleC::PeleC()
     mms_src_evaluated(false)
 #endif
 {
+  typical_values_chem.resize(NUM_SPECIES + 1);
```
marchdf: can you just set the size and default value in the declaration: `static amrex::Vector<amrex::Real> typical_values_chem_usr(NUM_SPECIES +1, 0);` And then remove both resize's (here and below)?

Source/PeleC.H
```diff
@@ -691,6 +693,13 @@ protected:
...
+  static bool use_typ_vals_chem_usr;
```
marchdf: why are there two bools for this? Feels like one should be enough? I must be missing something

Source/React.cpp
```diff
@@ -1,5 +1,35 @@
...
+    for (int sp = 0; sp < NUM_SPECIES; sp++) {
+      amrex::Real rhoYs_min = S_new.min(UFS + sp);
+      amrex::Real rhoYs_max = S_new.max(UFS + sp);
+      typical_values_chem[sp] =
+        std::max(0.5 * (rhoYs_min + rhoYs_max), typical_rhoY_val_min);
+    }
+    typical_values_chem[NUM_SPECIES] = 0.5 * (minTemp + maxTemp);
+    reactor->set_typ_vals_ode(typical_values_chem);
+
+    amrex::Print() << "\n======Setting typical values for level " << level
```
marchdf: Can we remove the `====` stuff here and below?

Source/PeleC.H
marchdf: Oh I think I see. Instead of two bools, can we do one int? if 0, do nothing, if 1 do usr, if 2 do default ?

Source/PeleC.H
```diff
@@ -691,6 +693,13 @@ protected:
...
+  static amrex::Real typical_rhoY_val_min;
+  static int reset_typical_vals_int;
+  amrex::Vector<amrex::Real> typical_values_chem;
+  static amrex::Vector<amrex::Real> typical_values_chem_usr;
```
marchdf, line 700: feels like only one of these vectors is truly necessary. Depending on the type of typical values required, it will be filled differently? Suggesting this because these modes look exclusive so we can use the same vector.

Source/React.cpp
```diff
@@ -1,5 +1,35 @@
 #include "React.H"
 
+void
+PeleC::set_typical_values_chem()
+{
+
```
marchdf, line 6: Since this only gets called on init/restart. Why not just move the pp query's here and then move the new PeleC class members into here and just make them normal variables? That way everything is local to one spot.

marchdf (changes requested):
Good stuff! Mostly questions about stuff since I probably don't understand it all.

Source/PeleC.H
hsitaram: done

Source/PeleC.H
hsitaram: again, two vectors are needed with the assumption that typical_values_chem is needed at each level. Now that I think of it, I may not need class member variable. Will change this.

Source/React.cpp
hsitaram: It gets called after regrid and every reset_typ_vals_int interval.

Source/PeleC.cpp
hsitaram: There is a subtle difference. typical_values_chem is local to each level. My thought was to have different set of typvals for each level, the logic being a flame may well resolved in a finer level while it may be washed out at a coarser one which means the typvals are different? I removed it as a member now, it is a local varaible

Source/PeleC.H
hsitaram: We can which is more elegant, but I wanted to be consistent with peleLM is using. It has a single bool variable to enable typ_vals use. I am happy to change it though.

Source/React.cpp
hsitaram: done

Source/React.cpp
marchdf: ah ok nvm then

Source/PeleC.H
marchdf: Oh I see. Yeah then maybe we stay closer to LM? It's nice to have the same options...

marchdf (approved):

jrood-nrel (approved):

PR #376, opened 2021-11-16
Track min and max of important quantities

baperry2:
I made some modifications so the terminal output and datalog files track extrema of important quantities (e.g., species mass fractions) for debugging purposes and thought it might be generally useful. This tracking occurs at the same intervals as tracking integrated quantities but can be turned off independently. 

Some notes:
- I modified the function `maxDerive`, which in the previous implementation ignored coarse cells covered by fine cells (good), but assumed maxima were positive (bad, e.g. for eint). The new implementation does not ignore covered cells, but this should be okay - average down is always called before the tracking, so covered cells should never have more extreme values than the cells covering them.
- "massfrac" reports extrema across all mass fractions. The individual extrema for the species mass fractions corresponding to `pelec.fuel_name`, `pelec.ftrac_spec_name`, and a new input called `pelec.extrema_spec_name` are also reported if these input parameters are specified.

### DIFF
--- a/Source/Params/_cpp_parameters
+++ b/Source/Params/_cpp_parameters
@@ -396,10 +396,16 @@ track_grid_losses            int            0
 # how often (number of coarse timesteps) to compute integral sums (for runtime diagnostics)
 sum_interval                 int           -1
 
+# record extrema of certain variables along with the intergal sums
+track_extrema                int           1
+
+# track the extrema of this species in addition to other quantities
+extrema_spec_name            string	   ""
+
 # how often (simulation time) to compute integral sums (for runtime diagnostics)
 sum_per                      Real          -1.0e0
 
-# abort if we exceed CFL = 1 over the cource of a timestep
+# abort if we exceed CFL = 1 over the course of a timestep
 hard_cfl_limit               int           1
 
 # a string describing the simulation that will be copied into the
--- a/Source/SumIQ.cpp
+++ b/Source/SumIQ.cpp
@@ -2,6 +2,16 @@
 
 #include "PeleC.H"
 
+int PeleC::find_datalog_index(const std::string& logname)
+{
+  for (int ii = 0; ii < parent->NumDataLogs(); ii++) {
+    if (logname == parent->DataLogName(ii)) {
+      return ii;
+    }
+  }
+  return -1; // Requested log not found
+}
+
 void
 PeleC::sum_integrated_quantities()
 {
@@ -10,7 +20,7 @@ PeleC::sum_integrated_quantities()
   if (verbose <= 0) {
     return;
   }
-
+  
   bool local_flag = true;
 
   int finest_level = parent->finestLevel();
@@ -85,11 +95,10 @@ PeleC::sum_integrated_quantities()
         //               << '\n';
         amrex::Print() << "TIME = " << time << " FUEL PROD   = " << fuel_prod
                        << '\n';
-        amrex::Print() << "TIME = " << time << " TEMP        = " << temp
-                       << '\n';
 
-        if (parent->NumDataLogs() > 0) {
-          std::ostream& data_log1 = parent->DataLog(0);
+	const int log_index = find_datalog_index("datalog");
+        if (log_index >= 0) {
+          std::ostream& data_log1 = parent->DataLog(log_index);
           if (data_log1.good()) {
             const int datwidth = 14;
             if (time == 0.0) {
@@ -259,8 +268,10 @@ PeleC::monitor_extrema()
                          << std::endl;
         }
 
-        if (parent->NumDataLogs() > 1) {
-          std::ostream& data_log1 = parent->DataLog(1);
+	
+	const int log_index = find_datalog_index("extremalog");
+        if (log_index >= 0) {
+          std::ostream& data_log1 = parent->DataLog(log_index);
           if (data_log1.good()) {
             const int datwidth = 17;
             if (time == 0.0) {
--- a/Source/SumIQ.cpp
+++ b/Source/SumIQ.cpp
@@ -2,16 +2,6 @@
 
 #include "PeleC.H"
 
-int PeleC::find_datalog_index(const std::string& logname)
-{
-  for (int ii = 0; ii < parent->NumDataLogs(); ii++) {
-    if (logname == parent->DataLogName(ii)) {
-      return ii;
-    }
-  }
-  return -1; // Requested log not found
-}
-
 void
 PeleC::sum_integrated_quantities()
 {
@@ -20,7 +10,7 @@ PeleC::sum_integrated_quantities()
   if (verbose <= 0) {
     return;
   }
-  
+
   bool local_flag = true;
 
   int finest_level = parent->finestLevel();
@@ -96,7 +86,7 @@ PeleC::sum_integrated_quantities()
         amrex::Print() << "TIME = " << time << " FUEL PROD   = " << fuel_prod
                        << '\n';
 
-	const int log_index = find_datalog_index("datalog");
+        const int log_index = find_datalog_index("datalog");
         if (log_index >= 0) {
           std::ostream& data_log1 = parent->DataLog(log_index);
           if (data_log1.good()) {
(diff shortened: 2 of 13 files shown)


marchdf:
Hey Bruce, this is great! There are a bunch of things the linter picked up if you wouldn't mind fixing those. I left some other minor comments

Source/SumIQ.cpp
```diff
@@ -138,3 +138,137 @@ PeleC::sum_integrated_quantities()
 #endif
   }
 }
+
+void
+PeleC::monitor_extrema()
+{
+  BL_PROFILE("PeleC::sum_integrated_quantities()");
```
marchdf: rename the BL_PROFILE stuff

Source/SumIQ.cpp
```diff
@@ -138,3 +138,137 @@ PeleC::sum_integrated_quantities()
...
+
+  if (verbose <= 0) {
+    return;
+  }
+
+  bool local_flag = true;
+
+  int finest_level = parent->finestLevel();
+  amrex::Real time = state[State_Type].curTime();
```
marchdf: const this

Source/SumIQ.cpp
```diff
@@ -138,3 +138,137 @@ PeleC::sum_integrated_quantities()
...
+  amrex::Vector<std::string> extrema_vars = {
+    "density", "x_velocity", "y_velocity", "z_velocity",
+    "eint_e",  "Temp",       "pressure",   "massfrac"};
+
+  int nspec_extrema = 1;
+  if (!fuel_name.empty()) {
+    nspec_extrema += 1;
```
marchdf: `++` instead of `+=1` (here and below)

Source/SumIQ.cpp
```diff
@@ -138,3 +138,137 @@ PeleC::sum_integrated_quantities()
...
+    extrema_vars.push_back(flame_trac_name);
+  }
+  if (!extrema_spec_name.empty()) {
+    nspec_extrema += 1;
+    extrema_vars.push_back(extrema_spec_name);
+  }
+  auto nextrema = extrema_vars.size();
+  constexpr amrex::Real neg_huge = std::numeric_limits<amrex::Real>::lowest();
+  constexpr amrex::Real huge = std::numeric_limits<amrex::Real>::max();
+  amrex::Vector<amrex::Real> minima, maxima;
```
marchdf: the next four or five lines can just be put in the construction: `minima(nextrema, huge)`, basically the same thing for `maxima` too.

Source/SumIQ.cpp
```diff
@@ -138,3 +138,137 @@ PeleC::sum_integrated_quantities()
...
+  minima.resize(nextrema);
+  maxima.resize(nextrema);
+  for (int ii = 0; ii < nextrema; ++ii) {
+    maxima[ii] = neg_huge;
+    minima[ii] = huge;
+  }
+
+  for (int lev = 0; lev <= finest_level; lev++) {
+    PeleC& pc_lev = getLevel(lev);
+    for (int ii = 0; ii < nextrema - nspec_extrema; ++ii) {
```
marchdf, line 188: why `ii` and not just `i`?

Source/SumIQ.cpp
```diff
@@ -138,3 +138,137 @@ PeleC::sum_integrated_quantities()
...
+void
+PeleC::monitor_extrema()
+{
+  BL_PROFILE("PeleC::sum_integrated_quantities()");
+
+  if (verbose <= 0) {
+    return;
+  }
+
+  bool local_flag = true;
```
marchdf: const this?

Source/SumIQ.cpp
```diff
@@ -138,3 +138,137 @@ PeleC::sum_integrated_quantities()
...
+        maxima[ii], pc_lev.maxDerive(extrema_vars[ii], time, local_flag));
+      minima[ii] = amrex::min(
+        minima[ii], pc_lev.minDerive(extrema_vars[ii], time, local_flag));
+    }
+
+    // Handle species seperately
+    auto mf = derive("massfrac", time, 0);
+    BL_ASSERT(!(mf == nullptr));
+    int idx_massfrac = nextrema - nspec_extrema;
+    bool local = true;
```
marchdf: const this? Or take the same as above `local_flag`?

marchdf (changes requested):

marchdf:
Is there a reason not to just output the extrema of all the mass fractions? I guess it can get verbose...

marchdf:
And can you post an example of the log file with these modifications? Curious to see what it looks like.

baperry2:
Here is what the output looks like:
Screen Shot 2021-11-17 at 9 34 23 AM

And here is the log file:
Screen Shot 2021-11-17 at 9 34 03 AM

jrood-nrel:
I think I would be fine replacing the integrated quantities with this.

shashankNREL:
@baperry2 : A small suggestion - can you either print on the screen as exponentials with only a few digits or print just the one decimal point. We perhaps don't need that much resolution in min and max. You can retain the full in the log file.

shashankNREL:
> @baperry2 : A small suggestion - can you either print on the screen as exponentials with only a few digits or print just the one decimal point. We perhaps don't need that much resolution in min and max. You can retain the full in the log file.

marchdf:
> I think I would be fine replacing the integrated quantities with this.

I want to keep the integrated quantities. I use them for things like TG and quick plotting of integral quantities.

marchdf:
> @baperry2 : A small suggestion - can you either print on the screen as exponentials with only a few digits or print just the one decimal point. We perhaps don't need that much resolution in min and max. You can retain the full in the log file.

I second this. Can we also make it look more like the other ones instead of whitespace before the name, make the whitespace between the name and the equal? Also, a space between `MIN/MAX` and `=` sign?

Source/SumIQ.cpp
baperry2: I dislike single character variable names and try to avoid them to make search, find+replace, etc, easier to use.

marchdf: lol

baperry2:
Okay, I think I addressed all of the above comments. I also added tracking for the sum of mass fractions (actually sumY - 1 to make any discrepancy easier to track without requiring crazy precision). And I added an option to print extrema for all mass fractions by setting `pelec.extrema_spec_name = ALL`.

Source/SumIQ.cpp
jrood-nrel: Is your editor the stickies app?

marchdf:
Excellent idea. Can you repost the new output here?

baperry2:
Log file:
Screen Shot 2021-11-17 at 1 26 58 PM

Output:
Screen Shot 2021-11-17 at 1 26 18 PM

(without the option to print all species to keep these reasonably sized)

marchdf:
Sweet. One last thing. Can you remove `TEMP` from the Integrated quantities ?

Source/Params/_cpp_parameters
```diff
@@ -396,10 +396,16 @@ track_grid_losses            int            0
...
+track_extrema                int           1
+
+# track the extrema of this species in addition to other quantities
+extrema_spec_name            string	   ""
+
 # how often (simulation time) to compute integral sums (for runtime diagnostics)
 sum_per                      Real          -1.0e0
 
-# abort if we exceed CFL = 1 over the cource of a timestep
+# abort if we exceed CFL = 1 over the course of a timestep
```
marchdf, line 408: coarse not course

jrood-nrel:
Do we need to list `TIME` before everything? Looks really redundant now.

drummerdoc:
Not required, but makes it easy to make plots using grep in the log

jrood-nrel:
Maybe we could just use `QUANTITY` or something.

baperry2:
@drummerdoc I've found having TIME in the output handy for the same reason. But the datalog file is there for plotting purposes (except on runs when you didn't set it up, which you sometimes still want to plot).

My preference would be to record datalog files by default.

I need to reshuffle the log file in some way because I noticed it conflicts with the prob log file for some of the test cases.

baperry2:
Ok, I removed `TEMP` but left `TIME` there. Also modified things so datalogs are specified by name instead of index to avoid potential overlap. I think this is ready to merge.

marchdf (approved):

marchdf:
there's even documentation!!

PR #382, opened 2021-12-01
Reduce scope of some variables

jrood-nrel:


Source/SumIQ.cpp
```diff
@@ -44,6 +43,7 @@ PeleC::sum_integrated_quantities()
   }
 
   if (verbose > 0) {
+    amrex::Real enstr = 0.0;
```
marchdf: can you just remove `enstr`? We don't use it anyway

Source/MOL.cpp
```diff
@@ -134,13 +134,12 @@ pc_compute_hyp_mol_flux(
...
         amrex::Real maxeigval = 0.0;
 
-        amrex::Real tmp0 = 0.0;
-        amrex::Real tmp1 = 0.0;
-        amrex::Real tmp2 = 0.0;
-        amrex::Real tmp3 = 0.0;
-        amrex::Real tmp4 = 0.0;
-
         if (!use_laxf_flux) {
+          amrex::Real tmp0 = 0.0;
```
marchdf: can we make this a single line? `amrex::Real tmp0 = 0.0, tmp1 = 0.0, ...`

marchdf (approved):

Issue #251, opened 2021-02-22
Variable boundary conditions on embedded boundaries


asmunder:
Opening an issue to discuss inhomogeneous boundary conditions on the embedded boundaries, as suggested by @drummerdoc .
This might be implemented in PeleLM by @esclapez first (https://github.com/AMReX-Combustion/PeleLM/issues/157) and then taken up into PeleC.

In principle there are two cases - one that is simple to implement, one that is more general:

1. varying wall temperature, possibly Neumann on some regions.
2. ability to set all components of the state on different regions.  You could have an inlet jet on this part, varying temperature on this other part, injecting fuel over here, etc. In principle arbitrarily complicated.

For case 1 I have something that seems to be working, which I'll describe in the next comment.
For case 2 I will leave it up for discussion.

asmunder:
### Case 1.
During initialization of PeleC, the temperature on the boundary is set to a uniform value in a sparse data structure of type `EBBndrySten`, in the following piece of code that is the same in `Source/PeleC_init_eb.cpp` or `SourceCpp/InitEB.cpp`:
 
      if (eb_isothermal && (diffuse_temp != 0 || diffuse_enth != 0)) {
          sv_eb_bcval[iLocal].setVal(eb_boundary_T, QTEMP);
      }

Replacing that code with the following will set the boundary temperature to be varying as a function of space.
The `USER_DEFINED_TEMPERATURE` function here takes `x,y,z` as arguments and returns temperature for the specific case.
There should be a generic version of this function that just returns `eb_boundary_T`, and the user could modify that as wanted.

- Where is a logical place to put the `USER_DEFINED_TEMPERATURE` function?
 
      if (eb_isothermal && (diffuse_temp != 0 || diffuse_enth != 0)) {
          int ivec = 0;
          Real dx = geom.CellSize(0);
          Real bcx, bcy, bcz;
          for (auto& sv_ebg : sv_eb_bndry_geom[iLocal]) {
            const IntVect& iv = sv_ebg.iv;
            bcx = geom.ProbLo()[0] + iv[0]*dx;
            bcy = geom.ProbLo()[1] + iv[1]*dx;
            bcz = geom.ProbLo()[2] + iv[2]*dx;
            sv_eb_bcval[iLocal](ivec,cQTEMP) = USER_DEFINED_TEMPERATURE(bcx, bcy, bcz);
            ivec++;
          }
      }

**Caveats with the above code:**
[...]

PR #391, opened 2021-12-14
Adding user defined, time varying EB boundary values.

marchdf:
This doesn't work perfectly with velocity. MOL would need to be aware
of it and it is not yet. Another branch has an attempt at this
functionality. Also, the EB tangential derivatives for velocity would
need to be updated.

Discussion relevant to #251


marchdf:
To be clear: I don't necessarily think this should be merged. But I am hoping we can start the discussion.

PR #399, opened 2022-01-10
Update to SUNDIALS 6

jrood-nrel:


marchdf (approved):
Other than the CI not passing, this looks fine ;)

Source/main.cpp
```diff
@@ -46,8 +48,10 @@ main(int argc, char* argv[])
 
   // Make sure to catch new failures.
   amrex::Initialize(argc, argv);
-#ifdef AMREX_USE_GPU
-  amrex::sundials::Initialize();
+// Defined and initialized when in gnumake, but not defined in cmake and
+// initialization done manually
+#ifndef AMREX_USE_SUNDIALS
+  amrex::sundials::Initialize(amrex::OpenMP::get_max_threads());
```
balos1, line 54: Why isn't `AMREX_USE_SUNDIALS` initialized when using CMake? It should be defined right here: https://github.com/AMReX-Codes/amrex/blob/ef9a37d6a14a704d59f105945c203da717595cb5/Tools/CMake/AMReXSetDefines.cmake#L170.

jrood-nrel: So for gnumake we can easily using `USE_SUNDIALS` and `make TPL`, but since I still want to use `add_subdirectory` for cmake rather than build sundials separately, we bypass the amrex sundials initialization and do it manually in PeleC. Does that make sense?

PR #406, opened 2022-01-31
Remove limiting of dcen in PLM

marchdf:
This was introduced through an email exchange with JBB. Follow-on
discussion almost 2y later with JBB and AA revealed that this
shouldn't actually have been introduced.


drummerdoc (approved):
Nailed it!

PR #409, opened 2022-02-08
Clean mass fractions in cut cells

marchdf:
This PR does a couple of things and is based on an idea that @esclapez prototyped:
- removes old `clean_massfrac` option that cleaned the mass fractions in `ctoprim`. We never use it and it seemed to do strange things to the flow
- introduces a new `eb_clean_massfrac` option (default to `true`): this resets the mass fractions *in cut cells only (and cells next to cut cells)* if they are out-of-bounds (`< tolerance` or `> 1 + tolerance`, where `tolerance` defaults to 0)

The code resets the mass fractions, density, momentum, internal energy and total energy (preserving kinetic energy) inside cut cells (and cells next to cut cells) and only in the hydro/diffusion routine, after the state redistribution has happened. It ends up modifying the divergence such that the new state has clean mass fractions in the cut cells. This implies that the `molSrc` resulting from this operation is adequately passed into the `react` call. 

I've tested this on the multi-species shock tube (EB-C8) and it removes all OOB mass fractions and the solution still looks good. Olga has tested the prototyped version of this that Lucas made and it helped her case.


jrood-nrel (approved):
Cool. Looks fine with me.

drummerdoc:
Seems ok, but I don't have the cycles this week to check carefully

hsitaram (approved):

PR #411, opened 2022-02-10
Fix rho after initial redistribution

marchdf:
In certain (rare) cases, the initial redistribution might change the `rho Y` at the EB. This ensures that `rho` is updated to reflect those changes (exactly like we do for the divergence of rho after state redistribution)

### DIFF
--- a/Source/PeleC.H
+++ b/Source/PeleC.H
@@ -566,6 +566,12 @@ public:
 
   // static EB helper functions called from outside class
   static bool ebInitialized();
+
+  void InitialRedistribution(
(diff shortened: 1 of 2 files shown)


jrood-nrel:
Do you just have to move https://github.com/AMReX-Combustion/PeleC/blob/9dee93fda4a3047b240c71d8b33abf49b378354c/Source/PeleC.H#L652 out of the `protected` list for CUDA?

marchdf:
You are right. I don't know why that works but I guess it does. Thanks!

PR #418, opened 2022-02-24
Add TurbInflow and example case

esclapez:


Source/BCfill.cpp
```diff
@@ -49,6 +49,13 @@ struct PCHypFillExtDir
       for (int n = 0; n < NVAR; n++) {
         s_int[n] = dest(loc, n);
       }
+#ifdef PELEC_USE_TURBINFLOW
+      if(iv[idir] == domlo[idir]-1){
+        for (int n = 0; n < NVAR; n++) {
+          s_ext[n] = dest(iv[0], iv[1], iv[2], n);
```
marchdf: can we just `dest(iv, n)` ?

Source/BCfill.cpp
```diff
@@ -60,6 +67,13 @@ struct PCHypFillExtDir
       for (int n = 0; n < NVAR; n++) {
         s_int[n] = dest(loc, n);
       }
+#ifdef PELEC_USE_TURBINFLOW
+      if(iv[idir] == domlo[idir]+1){
+        for (int n = 0; n < NVAR; n++) {
+          s_ext[n] = dest(iv[0], iv[1], iv[2], n);
```
marchdf: same here: can we just `dest(iv, n)` ?

Source/BCfill.cpp
```diff
@@ -73,6 +87,13 @@ struct PCHypFillExtDir
       for (int n = 0; n < NVAR; n++) {
         s_int[n] = dest(loc, n);
       }
+#ifdef PELEC_USE_TURBINFLOW
+      if(iv[idir] == domlo[idir]-1){
+        for (int n = 0; n < NVAR; n++) {
+          s_ext[n] = dest(iv[0], iv[1], iv[2], n);
```
marchdf: and here. And I guess the other spots ;)

Source/BCfill.cpp
```diff
@@ -93,9 +121,19 @@ struct PCHypFillExtDir
     // zlo and zhi
     idir = 2;
     if ((bc[idir] == amrex::BCType::ext_dir) && (iv[idir] < domlo[idir])) {
+
```
marchdf: remove newline

Source/BCfill.cpp
```diff
@@ -93,9 +121,19 @@ struct PCHypFillExtDir
...
       for (int n = 0; n < NVAR; n++) {
         s_int[n] = dest(iv[0], iv[1], domlo[idir], n);
       }
+
```
marchdf: remove newline

Source/BCfill.cpp
```diff
@@ -93,9 +121,19 @@ struct PCHypFillExtDir
...
+#ifdef PELEC_USE_TURBINFLOW
+      if(iv[idir] == domlo[idir]-1){
+        for (int n = 0; n < NVAR; n++) {
+          s_ext[n] = dest(iv[0], iv[1], iv[2], n);
+        }
+      }
+#endif
+
```
marchdf: remove newline

Exec/Make.PeleC
```diff
@@ -28,6 +28,12 @@ ifeq ($(USE_FORCING), TRUE)
   DEFINES += -DPELEC_USE_FORCING
 endif
 
+ifeq ($(USE_TURBINFLOW), TRUE)
```
marchdf: I don't know that we want another ifdef. @jrood-nrel what do you think? @esclapez what's the danger of always compiling it it?

Source/BCfill.cpp
```diff
@@ -145,10 +190,86 @@ pc_bcfill_hyp(
...
-  const ProbParmDevice* lprobparm = PeleC::d_prob_parm_device;
-  amrex::GpuBndryFuncFab<PCHypFillExtDir> hyp_bndry_func(
-    PCHypFillExtDir{lprobparm});
+  ProbParmDevice* probparmDD = PeleC::d_prob_parm_device; // probparm data for device
+  ProbParmDevice* probparmDH = PeleC::h_prob_parm_device; // host copy of probparm data for device
+  ProbParmHost* probparmH = PeleC::prob_parm_host;        // probparm data for host
+  constexpr int dim = AMREX_SPACEDIM;
+
+#ifdef PELEC_USE_TURBINFLOW
+  if (probparmH->do_turb) {
```
marchdf: I am thinking maybe we move this block and the one after this one to their own functions. Just to keep things more readable?

marchdf:
@esclapez thanks for doing this! I am happy to address my own comments, take care of formatting and that kind of stuff if you want to do other things.

marchdf:
I just went ahead and did that... hopefully you don't mind.

Source/BCfill.cpp
```diff
@@ -145,10 +189,94 @@ pc_bcfill_hyp(
...
+      amrex::Gpu::hostToDevice, probparmDH, probparmDH + 1, probparmDD);
+  }
+#endif
+
   amrex::GpuBndryFuncFab<PCHypFillExtDir> hyp_bndry_func(
-    PCHypFillExtDir{lprobparm});
+    PCHypFillExtDir{probparmDD});
   hyp_bndry_func(bx, data, dcomp, numcomp, geom, time, bcr, bcomp, scomp);
+
+#ifdef PELEC_USE_TURBINFLOW
```
marchdf: Why does this chunk of code exist? Seems like we are just setting some things to false? why is that?

drummerdoc: Probably not necessary, but is cleaning up after itself to return the state of the flags.  Yes, can remove

esclapez: We can indeed. I removed it from the LM implementation.

Exec/Make.PeleC
esclapez: I'm not sure there really is a danger. But if we skip those we'll need to always have a `turb` object in the `probparm` or find another mechanism to check if we want to inject turbulence or not.

marchdf: Ok yeah I see the problem... Do you think it would make sense to orchestrate the turbulence parm stuff outside of the prob stuff? It feels like if we have a bunch of these problems then they have a whole bunch of duplicate code.

Also I think we only need a host version of the turbulence parm stuff. So if we lift it out of the probparm then we can avoid all the GPU-CPU copies. Or am I missing something? All we need is `tp` right? And that can live outside the prob stuff.

marchdf: we can use `tp.turbinflow_initialized` to detect if the code should call the turbulence inflow. so we don't need `do_turb`

marchdf:
Needs https://github.com/AMReX-Combustion/PelePhysics/pull/232

marchdf:
Mohammad tested this out and it seems to work fine: 
Screen Shot 2022-02-27 at 12 11 25 PM

drummerdoc (approved):

PR #426, opened 2022-03-10
Bound initial small temperature values in pc_rst_int_e

jrood-nrel:
This avoid overflows with the Intel compiler. Also removed commented out code. I'm just making up `small_num` but at least it's still scaled by `numeric_limits`.

### DIFF
--- a/Source/Utilities.H
+++ b/Source/Utilities.H
@@ -48,7 +48,6 @@ pc_rst_int_e(
   }
   auto eos = pele::physics::PhysicsType::eos();
   if (allow_small_energy == 0) {
-    std::numeric_limits<amrex::Real>::epsilon() * 1e-100;
     const amrex::Real rhoInv = 1.0 / S(i, j, k, URHO);
     const amrex::Real Up = S(i, j, k, UMX) * rhoInv;
     const amrex::Real Vp = S(i, j, k, UMY) * rhoInv;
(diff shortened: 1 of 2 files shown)


Source/Utilities.H
```diff
@@ -117,8 +118,9 @@ pc_rst_int_e(
       if (
         S(i, j, k, UEINT) <
         (S(i, j, k, URHO) * std::numeric_limits<amrex::Real>::min())) {
-        // const amrex::Real eos_state_rho = S(i, j, k, URHO);
-        const amrex::Real eos_state_T = std::numeric_limits<amrex::Real>::min();
+        const amrex::Real small_num =
```
marchdf: can we pull the definition of `small_num` out of the if conditions here so we only have one declaration for all of these, instead of 3? Another thing we could do is put this into `Constants.H`. There's already a `small_num` there. Maybe a `real_small_num`? haha

jrood-nrel: Agreed.

marchdf (approved):

drummerdoc:
The only comment here about that is that when someone looks at the constants file, they have no clue where the small number plays a role.  Because of that, I'd rather declare the variable with a name that suggests what it is used for, then potentially have a few of them laying around for different purposes.  But, this might be cause I have a really bad memory.

jrood-nrel:
I just wanted to solve the overflows so I'm going to merge this and style will have to come later.

PR #427, opened 2022-03-10
Remove SYSTEM from certain include directories and add it to others

jrood-nrel:
Going to see what warnings appear when we remove these.


marchdf (approved):

jrood-nrel:
Hmm. I expected warnings. Guess we'll see what the dashboard does.

jrood-nrel:
Oh I still remove anything related to submodules with grep. That is why.

PR #428, opened 2022-03-11
Enabling initialization from a plot file

marchdf:
This PR allows for initialization from a plot file. Things to note:
- It uses `rho, u, T, Y` from a plot file to initialize a new state.
- The species in the new simulation will be taken from the plot file. Those that are not in the plot file will be set to zero. Those that are in the plot file but are not in the new simulation will be ignored (leading most probably to an error in species not summing to 1)
- It will sanitize the species mass fractions to ensure that they fall within the right bounds. 
- It will error out if the species are too far out of bounds
- This isn't quite like a restart. At the very least, the time stamp will restart at 0.

This is all explained in the documentation:
```  
   It is possible to initialize a simulation using a plot file
   (e.g. `pelec.init_pltfile = "plt00000"`). It uses :math:`\rho`,
   :math:`u`, :math:`T`, :math:`Y` from a plot file to initialize a
   new state. The species in the new simulation will be taken from the
   plot file. The species that are not in the plot file will be set to
   zero. The species that are in the plot file but are not in the new
   simulation will be ignored (leading most probably to an error in
   species not summing to 1). It is therefore assumed that the
   non-zero species in the plot file used to initialize the simulation
   form a subset of the species in the simulation. The code will
   sanitize the species mass fractions to ensure that they fall within
   the right bounds. It will error out if the species are too far out
   of bounds (i.e., too far below 0, too far above 1, not summing to
   1).
```

I tried restarting a simulation from the `air` mechanism using the `drm19` mechanism and it all looked fine.

### DIFF
--- a/Source/IO.cpp
+++ b/Source/IO.cpp
@@ -433,32 +433,6 @@ PeleC::setPlotVariables()
 
   bool plot_massfrac = false;
   pp.query("plot_massfrac", plot_massfrac);
-  //    if (plot_massfrac)
-  //    {
-  //	if (plot_massfrac)
-  //	{
-  //	    //
-  //	    // Get the species names from the network model.
-  //	    //
-  //	    for (int i = 0; i < NUM_SPECIES; i++)
-  //	    {
-  //		int len = 20;
-  //		Vector<int> int_spec_names(len);
-  //		// This call return the actual length of each string in "len"
-  //		get_spec_names(int_spec_names.dataPtr(),&i,&len);
-  //		char* spec_name = new char[len+1];
-  //		for (int j = 0; j < len; j++)
-  //		    spec_name[j] = int_spec_names[j];
-  //		spec_name[len] = '\0';
-  //		string spec_string = "Y(";
-  //		spec_string += spec_name;
-  //		spec_string += ')';
-  //		parent->addDerivePlotVar(spec_string);
-  //		delete [] spec_name;
-  //	    }
-  //	}
-  //    }
-
   if (plot_massfrac) {
     amrex::Amr::addDerivePlotVar("massfrac");
   } else {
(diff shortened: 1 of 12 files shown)


Source/IO.cpp
```diff
@@ -1194,3 +1195,102 @@ PeleC::writeSmallPlotFile(
...
+  // Read rho, u, temperature (required)
+  std::map<std::string, int> vars{
+    {"density", -1}, {"x_velocity", -1}, {"Temp", -1}};
+  for (auto& var : vars) {
+    var.second = find_position(plt_vars, var.first);
+    if (var.second == -1) {
+      amrex::Abort("Unable to find variable in plot file: " + var.first);
+    }
+  }
+  pltData.fillPatchFromPlt(lev, geom, vars["density"], URHO, 1, state);
```
baperry2: Another option would be to take the pressure from the plot file and use that to compute density. In an ideal world, these would be equivalent. But in a case where the various mass fraction modifications come into play, that might not be the case.

Source/IO.cpp
```diff
@@ -1194,3 +1195,102 @@ PeleC::writeSmallPlotFile(
...
+        sumY += sarr(i, j, k, UFS + n);
+      }
+
+      // If the sumY isn't too far from 1, renormalize
+      if (amrex::Math::abs(1.0 - sumY) < tol) {
+        for (int n = 0; n < NUM_SPECIES; n++) {
+          sarr(i, j, k, UFS + n) /= sumY;
+        }
+      } else {
+        amrex::Abort("Species mass fraction don't sum to 1");
```
baperry2: Can you add the actual sum to this error statement?

Source/IO.cpp
```diff
@@ -1194,3 +1195,102 @@ PeleC::writeSmallPlotFile(
...
+    state, [=] AMREX_GPU_DEVICE(int nbx, int i, int j, int k) noexcept {
+      auto sarr = sarrs[nbx];
+      amrex::Real sumY = 0.0;
+      const auto tol = constants::small_num();
+
+      for (int n = 0; n < NUM_SPECIES; n++) {
+        // if the species is not too far out of bounds, clip it
+        const auto mf = sarr(i, j, k, UFS + n);
+        if ((mf < 0.0) || (1.0 < mf)) {
+          if (((-tol < mf) && (mf < 0.0)) || ((1.0 < mf) && (mf < 1 + tol))) {
```
baperry2, line 1202: How about making `tol` something that is just defined for this function, or even a parmparsing it? For the use case of going between skeletal and reduced mechanisms, I could see it being useful to control this independently of whetever else `constants::small_num()` controls. I agree with this comment from @drummerdoc https://github.com/AMReX-Combustion/PeleC/pull/426#issuecomment-1064641187

baperry2 (commented):
This seems quite useful. Thanks Marc! IDK if this has been discussed, but  in addition to allowing us to go from nonreacting to reacting simulations, it should also allow us to go between skeletal and reduced mechanisms.

Source/IO.cpp
marchdf: Good idea. Added that.

Source/IO.cpp
marchdf: Good idea. I added that too.

baperry2 (approved):

Source/IO.cpp
marchdf: Bruce and I talked about this. And for now this PR is fine.

esclapez (approved):
Looks good Marc ! Note that this should be able handle initialization with a box distribution different than that of the pltfile.

PR #432, opened 2022-03-11
Chem model should really be Null for gammalaw

marchdf:
This will break the dashboard since I am basically removing a species. So fcompare will detect "variable not present" kind of thing.


baperry2 (approved):
This change makes sense to me.

jrood-nrel (approved):

jrood-nrel:
You can skip updating PelePhysics here.

PR #438, opened 2022-03-14
tag based on density ratio

baperry2:
This enables tagging based on the density ratio in adjacent cells (rather than the density difference). There are a few reasons for this. 

- The density ratio is nondimensional and therefore easier to interpret and not dependent on problem specifics. For example, one needs to change the density difference threshold based on temperature, pressure, and composition, and without knowing characteristic densities for a particular simulation it is hard to assess how much refinement a particular threshold for density difference will cause. 

- There is also a physics-based rationale for this criterion. The continuity equation is `div u = -1/rho Drho/Dt = - D(ln rho)/Dt`. Therefore, the direct physical effect of density change between cells is a function of the density ratio rather than the density difference. 

Draft for now because I only did very basic testing.

### DIFF
--- a/Source/Tagging.H
+++ b/Source/Tagging.H
@@ -9,6 +9,7 @@ struct TaggingParm
 {
   amrex::Real denerr = 1.0e10;
   amrex::Real dengrad = 1.0e10;
+  amrex::Real denratio = 1.0e10;
   amrex::Real presserr = 1.0e10;
   amrex::Real pressgrad = 1.0e10;
   amrex::Real velerr = 1.0e10;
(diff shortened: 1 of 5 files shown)


Source/Tagging.H
```diff
@@ -11,6 +11,8 @@ struct TaggingParm
   int max_denerr_lev = 10;
   amrex::Real dengrad = 1.0e10;
   int max_dengrad_lev = 10;
+  amrex::Real denratio = 1.0e10;
```
marchdf, line 12: we re-ordered these in a recent PR if you wouldn't mind following that

marchdf (approved):

olgadoronina (approved):

PR #441, opened 2022-03-15
Add readability-implicit-bool-conversion check

jrood-nrel:
I'm in support of this one, because it makes it clear we're treating ints as bools, but I understand skipping this one if it makes the code look undesirable.


marchdf:
So why don't we make some of these actual bools? `verbose`, `do_mol`, `do_react` etc

jrood-nrel:
I think I looked at that before, but not sure parmparse can do bools?

drummerdoc:
pp.val = t

PR #444, opened 2022-03-17
Turn on readability-implicit-bool-conversion check

jrood-nrel:
I think we need to convert some more.


Source/Advance.cpp
```diff
@@ -89,7 +89,7 @@ PeleC::do_mol_advance(
 #endif
 
   // Compute S^{n} = MOLRhs(U^{n})
-  if (verbose) {
+  if (verbose != 0) {
```
marchdf, line 92: I looked into doing `verbose` but it looks like we have different levels of verbosity. We have one check for `verbose > 1`.

marchdf (approved):

marchdf (approved):

PR #447, opened 2022-03-21
Remove EB ifdef

jrood-nrel:
In the effort to remove compile time options, here we commit to having the EB code always enabled. The main drawback to doing so is that we give up 1D ability in PeleC which is barely hanging on and only works in very limited situations anyway. I think removing the ifdef and 1D is worth more than keeping 1D because it simplifies several areas.


jrood-nrel:
Chart1
Here is a comparison running on the GPU. Two tests have a large slowdown we will probably need to investigate. I did three runs each and averaged them.

jrood-nrel:
cpu
CPU looks similar.

jrood-nrel:
After investigation, these tests just have higher initialization cost, and the timesteps in this branch look slightly faster.

jrood-nrel:
This is ready and there are no diffs between the stdout reported quantities.

jrood-nrel:
Screen Shot 2022-03-23 at 11 03 40 AM
I think we can solve the init cost for non-EB problems based on this profile.

esclapez:
Yep, this is an expensive piece of code. We might need a more clever way to do that.

marchdf (approved):

baperry2 (approved):
This dramatically improves code readability so I am all for it.

The slight increase in per timestep performance is surprising though. Any idea why that's happening (or maybe it's just noise)?

jrood-nrel:
I think it's noise. Without initialization increase (which I think @marchdf has since solved) in the graphs, runtimes look to be the same in general.

PR #450, opened 2022-03-23
Avoid some EB calculations when running non-EB problems

marchdf:
Needs #447


jrood-nrel (approved):

jrood-nrel:
new
Just thought I would confirm this. Red is these changes.

marchdf:
perfect! Thanks for confirming that this PR fixed those spikes.

PR #451, opened 2022-03-23
Fix turbinflow bug and use interior pressure to set jet inflow conditions

marchdf:


marchdf:
There was an actual bug in the inflow code. This:
```
 PeleC::turb_inflow.add_turb(
          bndryBoxLO, data, 0, geom, time, dir, amrex::Orientation::low);
```
Should be
```
 PeleC::turb_inflow.add_turb(
          bndryBoxLO, data, UMX, geom, time, dir, amrex::Orientation::low);
```
which means the current code was putting fluctuations in rho, umx, umy and not umx, umy, umz. So we were missing fluctuations in the z direction of velocity. @drummerdoc and Mohammad, my guess is that this drives the differences between LM and C. 

Current code:
Screen Shot 2022-03-30 at 9 49 14 AM

This PR:
Screen Shot 2022-03-30 at 10 52 49 AM

Ignoring differences in time stamps and scales, it is clear that there are now fluctuations in the z-velocity.

jrood-nrel (approved):

PR #453, opened 2022-03-23
Reorganize headers

jrood-nrel:
These are changes from running the include what you use tool where it explicitly includes exactly what headers are necessary for each file.


.github/workflows/ci.yml
```diff
@@ -16,7 +16,7 @@ jobs:
         source: './Source ./Exec'
         exclude: '.'
         extensions: 'H,h,cpp'
-        clangFormatVersion: 11
+        clangFormatVersion: 10
```
marchdf, line 19: boooo make this 13! ;)

Source/Utilities.H
```diff
@@ -1,11 +1,48 @@
...
+#include "AMReX_Reduce.H"           // for Sum
+#include "AMReX_Vector.H"           // for Vector
+
+// PelePhysics
+#include "Fuego.H"            // for Fuego
+#include "GammaLaw.H"         // for GammaLaw
+#include "mechanism.H"        // for NUM_SPECIES
+#include "PelePhysics.H"      // for PhysicsType
+#include "PhysicsConstants.H" // for Constants, Constants::RU
+#include "SRK.H"              // for SRK
```
marchdf, line 30: Feels weird to include these? Shouldn't we just be able to include EOS? Or some central PelePhysics header?

jrood-nrel: The idea is that everything is explicit in where it's coming from.

PR #454, opened 2022-03-24
Start building PelePhysics as libraries in CMake

jrood-nrel:
I will try to solve the DPC++ link failures on another machine before merging this.


CMakeLists.txt
```diff
@@ -131,15 +125,11 @@ endif()
 
 init_code_checks()
 
-#Turn on rpath stuff
 include(SetRpath)
 
-#Build pelec executables and link to amrex library
 add_subdirectory(Exec)
 
-if(PELEC_ENABLE_TESTS)
```
marchdf, line 140: Do we just always have testing on now? Fine by me

Exec/CMakeLists.txt
```diff
@@ -1,4 +1,2 @@
 add_subdirectory(RegTests)
-if(PELEC_ENABLE_TESTS)
-  add_subdirectory(UnitTests)
-endif()
+#add_subdirectory(UnitTests)
```
marchdf, line 2: Did you mean to disable these?

Exec/RegTests/EB-ConvergingNozzle/CMakeLists.txt
```diff
@@ -2,9 +2,4 @@ set(PELEC_ENABLE_PARTICLES OFF)
 set(PELEC_EOS_MODEL GammaLaw)
 set(PELEC_CHEMISTRY_MODEL Null)
 set(PELEC_TRANSPORT_MODEL Constant)
-
-#Generic stuff for every executable
-get_filename_component(DIR_NAME ${CMAKE_CURRENT_SOURCE_DIR} NAME)
-set(pelec_exe_name PeleC-${DIR_NAME})
-include(BuildPeleCExe)
-build_pelec_exe(${pelec_exe_name})
+include(BuildExeAndLib)
```
marchdf, line 5: Can we check to see if all this stuff is also done in the `Production` cases?

Exec/RegTests/MMS/prob.H
```diff
@@ -91,14 +91,7 @@ bcnormal(
...
   amrex::Real x0 = 0.0, x1 = 0.0, x2 = 0.0;
-
-  x0 = x[0];
-#if AMREX_SPACEDIM > 1
-  x1 = x[1];
-#endif
-#if AMREX_SPACEDIM > 2
-  x2 = x[2];
-#endif
+  AMREX_D_TERM(x0 = x[0];, x1 = x[1];, x2 = x[2];)
```
marchdf, line 94: yay much better!

marchdf (approved):

CMakeLists.txt
jrood-nrel: Yeah, I can't say I've ever turned it off and since CMake is mostly for testing the code, where I use GNUmake for running it, it seemed like we should always have testing on.

Exec/CMakeLists.txt
jrood-nrel: Yes, just for CI performance. Since we don't have unit tests anyway, just the framework, we could add them back if we ever write unit tests.

Exec/RegTests/EB-ConvergingNozzle/CMakeLists.txt
jrood-nrel: Do we want to build the Production cases? It would cancel out the speedup from this PR.

jrood-nrel: I guess maybe we should add these in. I can do so in a PR after this.

marchdf: I don't know that we need them to build in the CI. Maybe just be able to build them with cmake if we wanted

jrood-nrel:
This builds on the JLSE machine so the virtual environment is running out of memory during link here it seems.

PR #465, opened 2022-04-05
Remove NPASSIVE and explicitly deal with other variables

marchdf:
NPASSIVE and the passive mapping were a pain. If you wanted to do things differently for different members of the "passive" list then you were in trouble. This makes things explicit.

Things left to do:
- [x] check that I didn't break anything. Should be no diffs in 2D and in 3D.


ldowen (approved):
Looks good to me!

PR #469, opened 2022-04-07
Add linear variables

ldowen:
I am splitting the multiphysics modeling into a few smaller PRs to make it easier to review. This one adds linear variables which will be used for soot moments (where U=Q instead of U=rho Q). I have also removed some unused particle stuff. Almost all the particle stuff has changed anyway and will be replaced with a later PR.


Source/Advance.cpp
```diff
@@ -91,7 +91,7 @@ PeleC::do_mol_advance(
   for (int n = 0; n < src_list.size(); ++n) {
     if (
       src_list[n] != diff_src
-#ifdef AMREX_PARTICLES
+#ifdef SPRAY_PELEC
```
jrood-nrel: Can we use `PELEC_SPRAY` to be more consistent with others? Also, I assume the definition for this in the compiler will come in a later PR?

ldowen: Yes and yes.

CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ include(PeleCUtils)
 
 #General options for all executables in the project
 set(PELEC_DIM "3" CACHE STRING "Number of physical dimensions")
+option(PELEC_ENABLE_AMREX_PARTICLES "Enable particles in AMReX library" OFF)
```
jrood-nrel, line 14: Can we use `PELEC_ENABLE_SPRAY` here? I think it's up to you whether or not to consider the work as sprays or particles in general, but I would just want to be consistent.

jrood-nrel:
Looks fine with me, besides my nitpicks. Thanks for splitting up the PRs!

CMakeLists.txt
ldowen: Yeah. As of right now, PeleC only has spray particles in my specific branch. But PeleLM has separate particle functionality that I tried to preserve from things in IAMR. In general, I want to have a separate flag for sprays in case we want other particles. So I will change that flag to reflect that.

ldowen: Regarding this, there is a flag in SetAmrexOption.cmake `AMReX_PARTICLES` that I wanted to assign to the `PELEC_ENABLE_AMREX_PARTICLES`. Independently, there would a spray specific flag I would use. I'm not sure if that warrants having it in CMakeList.txt though?

jrood-nrel: Ok, how about we just use `PELEC_ENABLE_PARTICLES` for this and use that to turn on the AMReX particles?

jrood-nrel: Actually it's fine as it is. I think we will need to treat it like we did with the EB.

ldowen: Ok. I think I got it. I will do another commit soon. Thanks!

jrood-nrel: I think we will want to build AMReX always with particles enabled like we did with EB, so CMake only needs to build AMReX once. Then we can turn particles on or off in PeleC executables if that makes sense. This is only a CMake issue though.

Source/Advance.cpp
```diff
@@ -131,7 +131,7 @@ PeleC::do_mol_advance(
   for (int n = 0; n < src_list.size(); ++n) {
     if (
       src_list[n] != diff_src
-#ifdef AMREX_PARTICLES
+#ifdef PELEC_SPRAY
```
marchdf: Instead of having these ifdefs here, could we just always have `spray_src` in the enum and just keep the if condition? And then kinda like `les_src`, `spray_src` only gets `define`d if asked for? This would remove a whole bunch of ifdefs that are interleaved into if conditions.

Source/Advance.cpp
```diff
@@ -508,7 +443,7 @@ PeleC::do_sdc_iteration(
     }
   }
 
-#ifdef AMREX_PARTICLES
+#ifdef PELEC_SPRAY
```
marchdf, line 426: This ifdef isn't necessary if we are already testing `do_spray_particles`? I guess in general, we've been trying to not duplicate if conditions with a bool and an ifdef.

Source/Godunov.H
```diff
@@ -23,6 +23,21 @@ pc_cmpflx_passive(
                         : flxrho * 0.5 * (ql + qr);
 }
 
+AMREX_GPU_DEVICE
+AMREX_FORCE_INLINE
+void
+pc_cmpflx_linpassive(
```
marchdf: this is identical with `pc_cmpflx_passive`, just call that.

ldowen: Oh good catch!

Source/Godunov.H
```diff
@@ -55,6 +70,33 @@ pc_transdo_passive(
   qm(ivp1, qc) = compo / rrnew;
 }
 
+AMREX_GPU_DEVICE
+AMREX_FORCE_INLINE
+void
+pc_transdo_linpassive(
```
marchdf: So I think this is close enough to `pc_transdo_passive` that we should be able to combine them. If you add `rr` to the list of inputs, then you can pass in 1 and 0 for `flxrho` and things should just work.

Source/Godunov.H
```diff
@@ -95,6 +137,42 @@ pc_transdd_passive(
   qm(ivp2, qc) = compo / rrnewl + hdt * srcpass;
 }
 
+AMREX_GPU_DEVICE
+AMREX_FORCE_INLINE
+void
+pc_transdd_linpassive(
```
marchdf: Same comment with this one. Feels unnecessary

Source/Godunov.H
```diff
@@ -130,6 +208,36 @@ pc_transd_passive(
   qm(ivp1, qc) = compo / rrnew + hdt * srcpass;
 }
 
+AMREX_GPU_DEVICE
+AMREX_FORCE_INLINE
+void
+pc_transd_linpassive(
```
marchdf: and this

Source/Advance.cpp
ldowen: Removing those ifdefs would clean the code up quite a bit. My only concern is that the spray functionality depends on PeleMP, which I am not sure we want to have everything always depend on.

Source/IndexDefines.H
```diff
@@ -58,11 +58,22 @@
 
 #define UFS (UFA + NUM_ADV)
 #define UFX (UFS + NUM_SPECIES)
-#define NVAR (NTHERM + NUM_ADV + NUM_SPECIES + NUM_AUX)
+#define ULIN (UFX + NUM_AUX)
```
marchdf, line 61: Can you add documentation for these? Kinda like we do for `NUM_ADV` in the Docs?

Source/PeleC.H
```diff
@@ -34,7 +33,7 @@ enum StateType { State_Type = 0, Reactions_Type, Work_Estimate_Type };
 enum sources {
   ext_src = 0,
   forcing_src,
-#ifdef AMREX_PARTICLES
+#ifdef PELEC_SPRAY
```
marchdf: See comments above, would be nice to remove

Source/Advance.cpp
marchdf: Yeah that's a good point. I am just wondering if we could isolate the spray calls that rely specifically on PeleMP and ifdef those out?

Source/Godunov.H
ldowen: So you think I should add two inputs for those representing `qyp(iv, QRHO)` and `qym(ivp1, QRHO)` to be different rr values?

ldowen: I don't think it's the same otherwise. Since it you'd end up with 
`rr qyp(iv,qc) / rr - compn / rr` instead of `qyp(iv,qc) - compn`

ldowen: The `compn` would end up being divided by rr.

jrood-nrel (approved):

marchdf (approved):

PR #472, opened 2022-04-09
Add documentation with BC comparison for EB-ConvergingNozzle case

baperry2:
This PR modifies and adds documentation to the EB-ConvergingNozzle case. 

This case now compares a few different boundary condition specifictions, and demonstrates that using the interior pressure at the subsonic inlet and using a characteristic-based approach that fixes the pressure at the outlet is the preferred method (unless NSCBC boundaries are not implemented). 

The results now agree well with the analytical expressions for isentropic quasi-1D compressible flow.

This should help improve documentation as brought up here: https://github.com/AMReX-Combustion/PeleC/discussions/455


marchdf (approved):

marchdf:
Fantastic work, @baperry2 ! Thanks for doing that. There's just a couple of warnings to be addressed.

Exec/RegTests/EB-ConvergingNozzle/runstudy.bash
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
jrood-nrel: This looks specific to Eagle. Can we remove it?

baperry2: Yeah. Other cases have similar scripts in the Docs (e.g. https://amrex-combustion.github.io/PeleC/ebverification/C10/README.html#running-study) and I figured it might be handy to include the script in a file as well. But if that isn't desired, I'll just put it in the Docs like the other cases.

marchdf (approved):

jrood-nrel (approved):

PR #480, opened 2022-04-15
Fix compilation with PELE_USE_KLU = TRUE

bssoriano:
PeleC was not compiling with PELE_USE_KLU = TRUE. TPL lines were copied from LMeX Utils/make.PeleLMeX into Exec/Make.PeleC


esclapez (approved):
Can you remove the commented 2 lines starting with # Force TPL as ... ?
Otherwise, it looks good.

jrood-nrel (approved):

PR #497, opened 2022-05-09
Landon/fix mol passives

ldowen:
This fixes a bug that prevented passive variables like ADV, AUX, and LIN from being properly advected when the MOL method is used. The MOL slope routing mimics the method used for the tangential velocity component. The slopes are changed to use the same component ordering as the other primitive variable data.

### DIFF
--- a/Source/MOL.H
+++ b/Source/MOL.H
@@ -12,6 +12,23 @@
 AMREX_GPU_DEVICE
 AMREX_FORCE_INLINE
 void
+lax_passive_flux(
+  const amrex::Real ul,
+  const amrex::Real ur,
+  const amrex::Real rl,
+  const amrex::Real rr,
+  const amrex::Real ql,
+  const amrex::Real qr,
+  const amrex::Real maxeigval,
+  amrex::Real& flux_tmp)
+{
+  flux_tmp =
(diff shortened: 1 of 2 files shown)


marchdf:
This looks good to me. I tested it on the EB-C8 multispecies shock tube. This does mean that the species are advected slightly differently than the passive scalars but I think that's ok. @hariswaran can correct me ;)

hsitaram (commented):
looks good! I would have preferred not to use the super-diffuse Lax Friedrich style flux for the default scheme. can we do 
passive-scalar flux=(qint_iu>0)?leftstate * qint_iu:rightstate * qint_iu?

ldowen:
I have the passive variables only use the Lax method if all other variables are using the Lax method to be consistent. It is very diffusive but I assume it is better to have consistency between the variables. If the Lax solver isn't being used, then the `pc_cmpflux_passive` function is used.

hsitaram:
My mistake! I misread the cmpflx function. It is doing what I intended, using the sign of ustar.

hsitaram:
This is good to go. Thank you @ldowen

marchdf (approved):

PR #507, opened 2022-06-03
Reduce warnings

jrood-nrel:


Source/PeleC.cpp
```diff
@@ -787,7 +789,8 @@ PeleC::estTimeStep(amrex::Real /*dt_old*/)
 
     if (diffuse_temp) {
       auto const* ltransparm = trans_parms.device_trans_parm();
-      amrex::Real dt = amrex::ReduceMin(
+      amrex::Real dt = 0.0;
+      dt = amrex::ReduceMin(
```
marchdf, line 793: why can't we assign straight from the `ReduceMin`?

jrood-nrel: I think `ReduceMin` is too complicated for the tools to understand it is initializing the variable. We could suppress it, but I prefer this over suppressing it.

marchdf: gotcha.

PR #509, opened 2022-06-08
Update challenge problem setup

esclapez:


jrood-nrel (approved):

Exec/Production/ChallengeProblem/prob.H
```diff
@@ -415,7 +415,8 @@ struct MyProbTagStruct
            cos(prob_parm.cone_angle * constants::PI() / 180.0)) +
         t_loc * t_loc / (prob_parm.r_hole * prob_parm.r_hole);
 
-      if (elrSq < 1.2) {
+      // ProbHi[2] is 0.425 for this problem
+      if (elrSq < 1.2 && z > 0.41) {
```
marchdf, line 419: If this works, lets replace with a query to probhi so we don't get confused if we change geom later.

Source/InitEB.cpp
```diff
@@ -461,7 +461,8 @@ PeleC::initialize_signed_distance()
                         static_cast<amrex::Real>(parent->refRatio(ilev - 1)[0]),
                         static_cast<amrex::Real>(ilev));
     }
-    extentFactor *= std::sqrt(2.0); // Account for diagonals
+    extentFactor *=  std::sqrt(2.0)
```
marchdf: should this really just be `diagFac` from below?

marchdf: We could throw diagFac from here and below in something right after eb_detag_factor ppquery

Exec/Production/ChallengeProblem/prob.H
esclapez: I didn't want to change the function signature to not break other problems. Maybe passing `geomdata` instead of dx, problo and probhi separately ?

marchdf: we should totally pass in geomdata. Let's not worry about it with this PR.

Source/InitEB.cpp
esclapez: I'll do that

marchdf (approved):

PR #511, opened 2022-06-10
Enable use of async IO

marchdf:
We might go a different route later but for now...


marchdf:
Going to be useless because of #512

PR #526, opened 2022-07-05
Add reacting TG case

marchdf:
I did this work 2y ago and never committed it. Lucas was talking about running the same thing in LM so here are the files so he can get started quickly.


esclapez:
Thanks Marc ! I'll give it a try before approving, that'll help me debug the LM counterpart.

PR #533, opened 2022-07-21
Add ability to output using Ascent

jrood-nrel:


Source/PeleAscent.H
```diff
@@ -0,0 +1,12 @@
+#ifdef AMREX_USE_ASCENT
```
jrood-nrel: I don't think we need this. Also the compiler will get a warning of no symbols in the file if it's ever included. So this should go around any includes for this instead.

marchdf: done

marchdf (approved):

PR #571, opened 2022-11-21
Add amrex tagging utility

marchdf:


Docs/sphinx/InputFiles.rst
```diff
@@ -226,6 +226,22 @@ Additionally, tagging is supported for a user-specified species which can functi
...
+   # Tag inside a box and a velocity magnitude value
+   tagging.refinement_indicators = yLow magvel
+   tagging.yLow.in_box_lo = -0.1  -0.52  -0.85
+   tagging.yLow.in_box_hi =  3.1 -0.45    0.85
+
+   tagging.magvel.max_level     = 2
+   tagging.magvel.value_greater = 1.2e4
+   tagging.magvel.field_name    = magvel
+
+The following keys are implemented: `value_greater`, `value_less`, `vorticity_greater`, `adjacent_difference_greater`, `in_box_lo` and `in_box_hi` (to specify a refinement region), `max_level`, `start_time`, and `end_time`.
```
baperry2: Can you add something like `field_name` may be be any derived or state variable?

baperry2 (approved):
Looks good, thanks for doing this. Tagging syntax was one of the annoying differences between C and LM(eX), and this adds flexibility for C.

Docs/sphinx/InputFiles.rst
marchdf: Good idea. Done.

marchdf:
One thing I hesitated to do and decided not to: we could in theory remove all our own tagging functions and just use the amrex stuff added in this PR. But I was worried that would break a lot of input files and workflows. Maybe there's a way to keep the input file the same but redirect to this new way... but that might still lead to unexpected diffs. So I decided to leave the current way and just add to it.

baperry2:
I agree with that. I don't see any harm in keeping the old stuff around, and getting rid of it would break pretty much all input files. We could put something in the docs to encourage people to use the AMReX version as they create new cases though.

marchdf:
Good idea. I did that.

drummerdoc:
Hey!  That was MY stuff!  Not no stinking AMReX thing!

jrood-nrel (approved):

PR #574, opened 2022-12-03
Max wall time

baperry2:
Add an option to stop the simulation after a set amount of wall time, specified in hours.


jrood-nrel (approved):
I think this is fine. If we're looking to checkpoint before wall time is expired though, in the future we should probably trap the signal issued from the scheduler to checkpoint as the job is killed.

baperry2:
> If we're looking to checkpoint before wall time is expired though, in the future we should probably trap the signal issued from the scheduler to checkpoint as the job is killed.

That is pretty much the main intended use case here, and this is a simple if imperfect way to allow that. Definitely open to better solutions in the future.

PR #575, opened 2022-12-06
Update submodules and remove elixirs

marchdf:
Tasks:
- [x] check that this runs ok with MPI + CUDA


jrood-nrel:
I ran the challenge problem on Summit with this and it seems to work.

PR #587, opened 2022-12-23
Clean defines

marchdf:


marchdf:
Turns out this would trigger a whole bunch of work to do right. Namely we would have to catch all the `#if NUM_*` (and others of the sort) and replace with true if conditions. In which case I think the code stays compiled in and we might get other issues/lose out on other stuff. So I will be less ambitious about it all.

jrood-nrel (approved):

PR #589, opened 2023-01-11
Only compile Soot.cpp in CMake if soot is enabled

jrood-nrel:
Also fix plot number length in soot-flame.inp. The tests need a specific plot name.


ldowen:
@jrood-nrel This looks good to me. Should we also do the same thing for `Particles.cpp`?

ldowen (approved):

jrood-nrel:
Oops I had automerge on. Yeah I will do this for `Particle.cpp`.

PR #595, opened 2023-01-20
Update pelephysics and amrex, ignore soret coeffs

marchdf:
Need to merge https://github.com/AMReX-Combustion/PelePhysics/pull/349 first
Need to merge https://github.com/AMReX-Combustion/PeleMP/pull/8 first

### DIFF
--- a/Source/Diffusion.cpp
+++ b/Source/Diffusion.cpp
@@ -213,8 +213,9 @@ PeleC::getMOLSrcTerm(
         amrex::launch(gbox, [=] AMREX_GPU_DEVICE(amrex::Box const& tbx) {
           auto trans = pele::physics::PhysicsType::transport();
           trans.get_transport_coeffs(
-            tbx, qar_yin, qar_Tin, qar_rhoin, coe_rhoD, coe_mu, coe_xi,
-            coe_lambda, ltransparm);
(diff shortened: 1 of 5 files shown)


Source/Diffusion.cpp
```diff
@@ -213,8 +213,9 @@ PeleC::getMOLSrcTerm(
         amrex::launch(gbox, [=] AMREX_GPU_DEVICE(amrex::Box const& tbx) {
           auto trans = pele::physics::PhysicsType::transport();
           trans.get_transport_coeffs(
-            tbx, qar_yin, qar_Tin, qar_rhoin, coe_rhoD, coe_mu, coe_xi,
-            coe_lambda, ltransparm);
+            tbx, qar_yin, qar_Tin, qar_rhoin, coe_rhoD,
+            amrex::Array4<amrex::Real>(), coe_mu, coe_xi, coe_lambda,
```
baperry2, line 217: This could cause issues if use_soret is activated in the input file. How about adding an abort when initializing to ensure tparm.use_soret is false, to make sure no one inadvertently tries to use this capability while it is not supported in PeleC?

baperry2 (approved):

Source/Diffusion.cpp
marchdf: excellent idea, let me add that.

jrood-nrel (approved):

PR #597, opened 2023-02-06
Check if PELE_COMPILE_AJACOBIAN is set

gardner48:


jrood-nrel:
I guess I'm probably fine with this. I'm not sure what to do about the jacobian in general yet. My intention was to just add the define in the `GNUmakefile` when necessary until we narrow in on something: https://github.com/AMReX-Combustion/PeleC/blob/d044eec3a6ad65b31b9088673701f37ff0e06a2c/Exec/RegTests/MMS/GNUmakefile#L37

gardner48:
This seemed like the easiest solution for the moment but maybe it would be better to add this define based on other options like `PELE_USE_MAGMA` and `PELE_USE_KLU` (and possibly `PELE_USE_GINKGO` in the future)

jrood-nrel (approved):

jrood-nrel:
@gardner48 If you click "update branch" this should automerge.

gardner48:
@jrood-nrel looks like I need an approval for the CI workflow to run

jrood-nrel:
Ok let's try this again. Can you click the update branch button?

PR #601, opened 2023-02-14
Remove setting of device pointer for PMF data in soot reg test

ldowen:


jrood-nrel (approved):

marchdf:
@ldowen pelemp submod is being updated in #599 so you remove that from this PR.

PR #620, opened 2023-04-26
Rogue omp pragma near new ParallelFor

baperry2:
I'm pretty sure this is extraneous and just got left behind when the ParallelFors were switched over to multifab ParallelFors. With the OMP pragma, I get a nested/multiple MultiFab error when running with multiple OMP threads, but things work without it.


marchdf (approved):

marchdf:
Good catch!

PR #636, opened 2023-06-13
Reorganize MOL file

marchdf:
This should cause zero diffs because it is just a reorg. It brings relevant things closer together so that the next step is easier. I also rewrote a bunch of comments

### DIFF
--- a/Source/Diffusion.cpp
+++ b/Source/Diffusion.cpp
@@ -69,13 +69,6 @@ PeleC::getMOLSrcTerm(
   const amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> dxD = {
     {AMREX_D_DECL(dx1, dx1, dx1)}};
 
-  // BCs
-  const amrex::StateDescriptor* desc = state[State_Type].descriptor();
-  const auto& bcs = desc->getBCs();
-  amrex::Gpu::DeviceVector<amrex::BCRec> d_bcs(desc->nComp());
-  amrex::Gpu::copy(
-    amrex::Gpu::hostToDevice, bcs.begin(), bcs.end(), d_bcs.begin());
-
   auto const& fact =
     dynamic_cast<amrex::EBFArrayBoxFactory const&>(S.Factory());
   auto const& flags = fact.getMultiEBCellFlagFab();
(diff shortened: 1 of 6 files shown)


Source/Diffusion.cpp
```diff
@@ -73,14 +70,13 @@ PeleC::getMOLSrcTerm(
...
+  // BCs
   const amrex::StateDescriptor* desc = state[State_Type].descriptor();
   const auto& bcs = desc->getBCs();
   amrex::Gpu::DeviceVector<amrex::BCRec> d_bcs(desc->nComp());
   amrex::Gpu::copy(
     amrex::Gpu::hostToDevice, bcs.begin(), bcs.end(), d_bcs.begin());
 
-  // Fetch some gpu arrays
   prefetchToDevice(S);
   prefetchToDevice(MOLSrcTerm);
```
jrood-nrel: I never see these in other applications. Do we know if they even do anything?

marchdf: Yeah I have no idea. I think this is a relic. It also is only used for managed mem: https://github.com/AMReX-Codes/amrex/blob/96b811d2cb9f24e87428e4ebc9904b4dcdf73e7f/Src/Base/AMReX_BaseFab.H#L1689. 

So I will just remove these.

Source/Diffusion.cpp
```diff
@@ -31,15 +31,16 @@ PeleC::getMOLSrcTerm(
```
baperry2, line 29: I think the part about face-centering in steps 1-2 above is not accurate, can that be updated

Source/Diffusion.cpp
```diff
@@ -73,14 +70,13 @@ PeleC::getMOLSrcTerm(
   const amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> dxD = {
     {AMREX_D_DECL(dx1, dx1, dx1)}};
 
-  // Grab the BCs
+  // BCs
```
baperry2: I think these are only used for EB redistribution so this could be moved down to the EB section

baperry2 (approved):

Source/Diffusion.cpp
marchdf: good catch, updated

Source/Diffusion.cpp
marchdf: another good catch, moved them down to that section

jrood-nrel (approved):

PR #640, opened 2023-06-15
Make tangential velocity weights  const on gpu instead of thread private

marchdf:


marchdf:
This is ever so slightly faster (0.16% in mol src term)
```
PeleC::getMOLSrcTerm()                             200      24.81      24.81      24.81  42.47%
```
vs current version:
```
PeleC::getMOLSrcTerm()                             200      24.85      24.85      24.85  42.17%
```

Source/Diffterm.H
```diff
@@ -11,15 +11,24 @@
...
+#ifdef AMREX_USE_GPU
+AMREX_GPU_CONSTANT const amrex::Real d_weights[3] = {0.0, 1.0, 0.5};
+#endif
+const amrex::Real h_weights[3] = {0.0, 1.0, 0.5};
+
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE amrex::Real
+weight(const int n)
+{
+#if AMREX_DEVICE_COMPILE
+  return d_weights[n];
```
jrood-nrel: Can it just return this with no ifdef?

marchdf: yes of course. I was being dumb and making this HOST_DEVICE. This is just a device function...

jrood-nrel (approved):

PR #646, opened 2023-06-27
Update PeleMP submodule

ldowen:
Updated the MP module to include fixes for things that show up with HIP but not CUDA for some reason.


jrood-nrel (approved):

jrood-nrel:
Yeah it looks like rocm 5.6.0 is having trouble with the use of `pow` in PeleMP. I'm not sure what to do about it yet.

jrood-nrel:
Actually since they are constants maybe you can just use constexpr or something for this struct: https://github.com/AMReX-Combustion/PeleMP/blob/40745796328eb3f03f4a5d9e5017a5bf62d73831/Source/Soot_Models/Constants_Soot.H#L53 .

PR #647, opened 2023-07-03
Redistribution move 

baperry2:
Bring PeleC up to date with AMReX changes moving redistribution from AMReX-hydro to main AMReX. As a result, PeleC in its present form does not rely on AMReX-hydro

https://github.com/AMReX-Codes/amrex/pull/3378
https://github.com/AMReX-Codes/AMReX-Hydro/pull/109


baperry2:
There was an issue with the cmake build using clang introduced by this amrex change (https://github.com/AMReX-Codes/amrex/pull/3309) where the `target_compile_options` function in `SetAmrexCompileFlags.cmake` doesn't work with an alias. The last commit should work around the issue, but IDK if there is a better way

marchdf (approved):

jrood-nrel:
I'm not convinced we won't need AMReX-Hydro in the future.

baperry2:
I can roll back the commits on removing the submodule, but we could also just add it back if the need arises later. Up to you.

jrood-nrel (approved):

PR #648, opened 2023-07-09
adding cavity flame input files

hsitaram:
1) added case files from PeleC paper:  Sitaraman et al. "Adaptive mesh based combustion simulations of direct fuel injection effects in a supersonic cavity flame-holder." Combustion and Flame 232 (2021): 111531.

2) removed extrusion step in extruded triangles. The triangle planes are already extruded.


marchdf (commented):
Can you remove the PelePhysics submod update? Otherwise looks good

hsitaram:
fixed the pelephysics update. although it says "changed 0 files", not sure what that means.. I never get submodules right

marchdf (approved):

baperry2:
Looking into some issues with the initialization, which appears to cause issues with excessive expansion near the leading edge of the cavity particularly on coarser grids

baperry2 (approved):
I added some different initialization options that appear to make the initial transient more stable. With these options, everything is stable even on a 128x32x8 grid with no refinement.

jrood-nrel:
I tried removing the submodule update on this branch, and I think this branch's history has a problem. I think we should play it careful with the submodule history so I made another PR for this https://github.com/AMReX-Combustion/PeleC/pull/649 .

PR #661, opened 2023-07-25
Remove track grid loses

marchdf:
This wasn't being used.


jrood-nrel (approved):

baperry2 (approved):
Yeah I noticed this as well, thanks for removing. No point in keeping this around if it not being used. Even if it were fully implemented, I don't think it would provide anything that you couldn't get by comparing sequential steps in the SumIQ output

PR #662, opened 2023-07-25
add BC checks in Godunov cmpflx

baperry2:
Bring in BC checks for walls and symmetry BCs from CAMR (https://github.com/AMReX-Codes/CAMR/pull/10).

To be added before merging this PR:
- [x] Simple test case demonstrating mass conservation

Future changes that will not be implemented in this PR:
- This PR does not yet implement the changes for artificial viscosity at NoSlip boundaries, so in PeleC NoSlip boundaries may cause flow features to form in physically inviscid simulations due to the artificial viscosity. 
- Default user BCs to noslip walls with true reflect_odd for velocity
- Modify treatment for user BCs in Godunov


marchdf (approved):

drummerdoc:
true reflect_odd for viscosity?  Not sure I get the meaning of that, or where that's done in the PR

baperry2:
Sorry that was a typo, it's been fixed to "true reflect odd for velocity". Right now, users have access to only the 1st interior cell when setting conditions in bcnormal, the plan is to give them both that and the reflected state (e.g., 3rd interior cell for the 3rd ghost cell) to enable better wall-type BCs to be set through bcnormal. But that's mostly unrelated to the contents of this PR.

This PR just makes sure that the appropriate wall BCs if applicable are applied when computing Godunov fluxes at the n+1/2 face states. This brings PeleC in line with CAMR.

baperry2:
@jrood-nrel I've added a new test case to verify mass conservation, can you double check how that is implemented?

jrood-nrel (approved):

PR #665, opened 2023-07-26
Change bcnormal interface to allow for better reflections at walls

baperry2:
The existing `bcnormal` interface gives the user the 1st interior cell only to set any ghost cell. That means that if using bcnormal to specify something like a wall, only the first ghost cell is a true reflection. This PR updates the interface so s_ext is prepopulated with true reflected state (even for most vars, odd for velocity, so corresponding to a NoSlipWall) before calling bcnormal. This gives the user the ability to better specify walls through bcnormal (for domain faces that are partially a wall and partially inflow). If a user wishes to have an inflow on part of a face and a NoSlipWall on the rest, they now need only to modify s_ext on the part of the face corresponding to the inflow.

The `MassCons` test case is modified to show examples of using bcnormal to specify wall BCs. 

Previously, s_ext was used to pass velocity fluctuations from turbulent inflows where applicable. However, that data structure is now used for the reflected state, so the bcnormal interface has to be modified to pass an additional array of the turbulent velocity fluctuations. **Therefore, this PR will break the `prob.H` for existing cases.** Users encountering compile errors after this PR should modify the definition of bcnormal in their `prob.H` for the new interface, as well as make the appropriate changes if defining any walls through bcnormal, as demonstrated in the MassCons case.

For now, besides the MassCons case, all other cases have had their bcnormal interface adjusted, but the function itself not modified. Therefore, this PR should not introduce any diffs. Once that has been confirmed, I will add in the changes to the bcnormal functions where relevant.


marchdf (approved):

baperry2:
I've modified bcnormal in the production cases to take advantage of s_ext being pre-populated as a wall. This should be ready to go now.

marchdf (approved):

PR #668, opened 2023-07-31
Add plm iorder flag back to PLM

marchdf:
Also snuck in a readme update


jrood-nrel (approved):

Source/PeleC.cpp
```diff
@@ -336,6 +336,13 @@ PeleC::read_params()
     amrex::Error("PeleC::ppm_type must be 1 (PPM) to use WENO method");
   }
 
+  if (do_hydro) {
+    if (do_mol || ppm_type == 0) {
+      if ((plm_iorder != 1) && (plm_iorder != 2) && (plm_iorder != 4))
+        amrex::Error("PeleC::plm_iorder must be 1, 2, or 4");
```
baperry2: I think for MOL we only really support 1 or 2, so there should be separate checks here for MOL and PLM

baperry2 (changes requested):

Source/PeleC.cpp
marchdf: good idea

Source/PeleC.cpp
```diff
@@ -337,9 +337,14 @@ PeleC::read_params()
...
 
   if (do_hydro) {
-    if (do_mol || ppm_type == 0) {
-      if ((plm_iorder != 1) && (plm_iorder != 2) && (plm_iorder != 4))
-        amrex::Error("PeleC::plm_iorder must be 1, 2, or 4");
+    if (ppm_type == 0) {
+      if ((plm_iorder != 1) && (plm_iorder != 2) && (plm_iorder != 4)) {
+        amrex::Error("PeleC::plm_iorder for ppm_type = 0 must be 1, 2, or 4");
+      }
+    } else if (do_mol) {
```
baperry2: Shouldn't `do_mol` be the `if` and `ppm_type` be the `else if` here?

marchdf: nice. good point. I think I fixed it?

baperry2 (approved):

PR #673, opened 2023-08-02
Reduce order at user BCs for Godunov

baperry2:
In bcnormal, users specify boundaries at step N, but the Godunov fluxes are computed at N+1/2. That leads to an inconsistency. This PR addresses that by computing fluxes for any boundary specified through bcnormal using the data only at step N, with face data always obtained with PLM at these boundaries. This does not affect other types of boundaries.


Source/Godunov.cpp
```diff
@@ -376,13 +410,37 @@ pc_umeth_3D(
...
   amrex::ParallelFor(zfxbx, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
-    pc_cmpflx(i, j, k, bclz, bchz, dlz, dhz, qm, qp, flx3, q3, qaux, cdir);
+    pc_cmpflx(
+      i, j, k, bclz, bchz, dlz, dhz, qm, qp, flx[2], qec[2], qaux, cdir);
   });
 
+  // Fix bcnormal boundaries - always use PLM and don't do N+1/2 predictor
+  // because the user specifies conditions at N
+  // bcnormal used for "Hard" (Inflow=1) and "UserBC" (6)
+  for (int idir = 0; idir < 3; idir++) {
```
marchdf: AMREX_SPACEDIM here?

Source/Godunov.cpp
```diff
@@ -376,13 +410,37 @@ pc_umeth_3D(
...
+    if (bclo[idir] == Inflow || bclo[idir] == 6) {
```
marchdf: instead of `6`, I think we have a variable to check that? Or maybe we dont...

Source/Godunov.cpp
```diff
@@ -510,7 +563,8 @@ pc_umeth_2D(
   amrex::ParallelFor(
     yflxbx, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
       pc_cmpflx(
-        i, j, k, bcly, bchy, dly, dhy, qymarr, qyparr, fyarr, q2, qaux, cdir);
+        i, j, k, bcly, bchy, dly, dhy, qymarr, qyparr, fyarr, gdtemp, qaux,
```
marchdf: did you mean to switch from q2 to gdtemp?

Source/Godunov.cpp
baperry2: Will do. This is in pc_umeth_3d so it should only be called when AMREX_SPACEDIM==3, but this would be more clear.

Source/Godunov.cpp
baperry2: I don't believe we do. The BC defines come from AMReX and they don't have a UserBC: https://github.com/AMReX-Codes/amrex/blob/e7d3d5a4fad6559ada6bfaa011438f9e7e317ca7/Src/Base/AMReX_BC_TYPES.H#L103

We may want to have our own defines like ERF: https://github.com/erf-model/ERF/blob/9ea6df90fc497bde2bc2a2e538480d4792507b9c/Source/IndexDefines.H#L159

Source/Godunov.cpp
baperry2: Yeah, although I will have to double check because I'm seeing unexpected errors. But in 3D all the equivalent calls put this into a temporary fab because it shouldn't be needed

baperry2: Turns out that was the issue, good catch

Source/Godunov.cpp
baperry2: Added an enum for this.

baperry2:
@jrood-nrel can you take a look at the test that's failing? I haven't seen something fail there before.

jrood-nrel:
I think the Github virtual environments may have been updated. It looks like we mostly have the option that would fix this, which is `-flto=auto` in the compiler lines, but we don't have it everywhere necessary. We can safely suppress it, which I did in #676 .

marchdf (approved):

PR #679, opened 2023-08-17
Fix reflux factor for MOL

marchdf:
This fixes mass conservation errors in MOL when there are multiple levels. Can someone tell me why this is not correct and why the previous way is correct?!


baperry2:
Can you add in the test that will break if this is wrong to this PR?

marchdf:
good idea, I tried to mimic what you did for masscons

Exec/RegTests/Sod/masscons.py
baperry2, line 1: Rather than creating a duplicate file, how about moving this file to `PeleC/Tests` similar to `test_second_order.py`.

Tests/CMakeLists.txt
```diff
@@ -224,6 +224,7 @@ add_test_r(hit-2 HIT)
 add_test_r(hit-3 HIT)
 add_test_r(sod-1 Sod)
 add_test_r(sod-2 Sod)
+add_test_rv(sod-3 Sod)
```
baperry2, line 227: How about converting any test without inflow/outflow to an `add_test_rv`, for example the HIT cases? Probably could add some of the EB ones as well.

baperry2 (approved):

Exec/RegTests/Sod/masscons.py
marchdf: excellent idea, done.

Tests/CMakeLists.txt
marchdf: Great, I think I did that now.

baperry2:
Looks like EB-C9 and EB-C12 just need to have datlog set in the inputs file; ChannelFlow conserves mass but fails on energy because of the velocity forcing.

marchdf:
Thanks for figuring that out!

PR #682, opened 2023-08-18
Make the tagging aware of EBs

marchdf:
Ignore data from covered cells in the EBs. Previously, even without tagging the EB, you would get something like this if you have a density or pressure grad tagging:
Screen Shot 2023-08-18 at 09 59 03

Basically, the tagging would look at covered cells and detect a gradient that needed to be refined. Now we ignore covered cell data for tagging. So you get this:
Screen Shot 2023-08-18 at 09 56 57

This should be merged after #678

### DIFF
--- a/Source/Tagging.H
+++ b/Source/Tagging.H
@@ -54,7 +54,7 @@ tag_error(
   const int i,
   const int j,
   const int k,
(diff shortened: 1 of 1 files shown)


baperry2 (approved):

jrood-nrel (approved):
Excellent, it makes the circle worse.

Issue #502, opened 2022-05-22
NoSlipWall in pelelm and pelec


czc-zju:
The following two cases are calculated by PeleLM and PeleC. In the results of PeleLM, we can see clear boundary layers, the velocity is very small, but the result of PeleC has no boundary layer, and the velocity on the vertical line is equal, why?
![图片1](https://user-images.githubusercontent.com/81547457/169676330-4a3d2c77-66bf-4455-9945-e3f508e7dbf7.png)

marchdf:
Please post your PeleC input file. Thanks!

czc-zju:
[example_inp.txt](https://github.com/AMReX-Combustion/PeleC/files/8750622/example_inp.txt)

czc-zju:
> 请张贴您的 PeleC 输入文件。谢谢！

I have tried many cases, but this problem can't be solved. 
Do I need to provide any other files?

marchdf:
Thanks for providing the input file. The first thing I see is that you have `pelec.diffuse_vel = 0`, which means that velocity diffusion is turned off, which means you have slip walls and not "no slip" walls, which would give a result as you see in PeleC. Switch it to `pelec.diffuse_vel = 1`. If this is multi-species, you should also turn on species diffusion.

drummerdoc:
Oops, switch to  pelec.diffuse_vel = 1 :)

czc-zju:
Thanks a million.

czc-zju:
I set NoSlipWall in my case in peleC, and I also set it as an isothermal wall (the following is my setting process), but the isothermal wall is not displayed in the figure. Is there something wrong with my method?
![2022-05-25_225810](https://user-images.githubusercontent.com/81547457/170296240-e302c9ed-696f-4a4d-84ea-ccf39fe336cf.png)

   _if (idir == 1 )  {
    auto eos = pele::physics::PhysicsType::eos();
    amrex::Real massfrac[NUM_SPECIES] = {0.0};
    for (int n = 0; n < NUM_SPECIES; n++)
    massfrac[n] = s_int[UFS + n] / s_int[URHO];
    const amrex::Real T = 500;
    amrex::Real eint=0.0;
    eos.RTY2E(s_int[URHO], T, massfrac, eint);
    s_ext[URHO] = s_int[URHO];
    s_ext[UMX] = 0.0;
    s_ext[UMY] = 0.0;
    s_ext[UMZ] = 0.0;
[...]

Issue #265, opened 2021-04-14
heat condution test fails


Toby-Yang:
has anyone tested the 1d heat conduction problem? I found PeleC failed to get accurate temperature distributions 
![图片1](https://user-images.githubusercontent.com/41934030/114675989-6c449f80-9d3b-11eb-9033-9b964eebae6c.png)

hsitaram:
can you give us a little more specifics for this case? Specifically about the bcnormal function you are using to impose the isothermal wall. An earlier closed issue  #263 may answer some of the issues you may be facing.

Toby-Yang:
thanks for you answer. I test a new case using NoSlipWall, the temperature distributions is  not agree with the results obtained with CHEMKIN package. So I think bcnormal function is  not not relevant to this issue

hsitaram:
How are you imposing the fixed temperature at the wall without bcnormal? are you using an EB?

ll-yang:
The new case use NoSlipWall  at left and right boundary, the initial temperature is a Gauss distribution and the gas is pure N2 at 1atm. this figure is a comparision. 0 microSecond is the initial temperature distribution. dashed line is the PeleC results
![无标题](https://user-images.githubusercontent.com/41934030/114907123-532c1380-9e4d-11eb-985a-44ba113c8b39.png)

hsitaram:
Feels like an issue with transport coefficients. Can you share your case with us? What chemistry_model are you using?

ll-yang:
I use the LiDryer model. The code  below is modified from PMF case. 

input:

```
# ------------------  INPUTS TO MAIN PROGRAM  -------------------
stop_time = 0.005
#max_step = 2

# PROBLEM SIZE & GEOMETRY
geometry.is_periodic = 0 1
geometry.coord_sys   = 0  # 0 => cart, 1 => RZ  2=>spherical
geometry.prob_lo     =   0.0        0.0   
geometry.prob_hi     =   2.0        0.01  
amr.n_cell           =   400        2 

#pelec.Riemann    = 0     # 0: HLL,  1: JBB,  2: HLLC
# >>>>>>>>>>>>>  BC KEYWORDS <<<<<<<<<<<<<<<<<<<<<<
# Interior, UserBC, Symmetry, SlipWall, NoSlipWall
# >>>>>>>>>>>>>  BC KEYWORDS <<<<<<<<<<<<<<<<<<<<<<
[...]

Issue #263, opened 2021-03-30
EB Isothermal BC for a supersonic channel case (@Mach 2.0)


RSuryaNarayan:
Hello everyone, 
I am trying to simulate supersonic flow inside a channel using PeleC. To do so I have taken the ```EB-Bluffbody``` case in the ```Exec/RegTests``` folder and made the following changes to the ```inputs.3d``` file 

1. Made changes to the domain
```
geometry.prob_hi     =  15.0  6.0  1.5
amr.n_cell           =  160   64   16
```
2. Modified BCs using:
```
pelec.lo_bc       =  "Hard"     "NoSlipWall" "Interior"
pelec.hi_bc       =  "FOExtrap" "NoSlipWall" "Interior"
```
3. Made changes to the EB to cover the entire domain
```
eb2.use_eb2 = 1
eb2.geom_type = box
eb2.box_lo = 0.0 0.0 0.0
eb2.box_hi = 15.0 6.0 1.5 
eb2.box_has_fluid_inside = 1
ebd.boundary_grad_stencil_type = 0
```
4. Changed the problem parameters 
```
prob.p = 1013250.0
prob.rho = 0.00116
prob.vx_in =  70000.0
prob.vy_in =  0.0
prob.Re_L = 625.0
prob.Pr = 0.7
```
Also since I want the walls to be isothermal at 300K I left the 
```
pelec.eb_boundary_T =300
pelec.eb_isothermal =1
```
as is. My question is, is this methodology right for the problem at hand or I need to modify something more to run this case as intended? The results obtained from the above inputs file look something like so: 
![Screenshot from 2021-03-30 15-57-26](https://user-images.githubusercontent.com/52670594/112975011-c60c6d80-9170-11eb-8fad-51de23bac2e4.png)
Here are the contours of the Mach number: 
![Screenshot from 2021-03-30 16-01-39](https://user-images.githubusercontent.com/52670594/112975510-619dde00-9171-11eb-964f-ec947157ea0b.png)

There is a nice shock wave and its reflections, however, the walls don't seem to maintain the 300K isothermal boundary condition, rather the temperatures vary along the length of the channel. Hence the whole question on the methodology behind the implementation. Thanks in advance.

marchdf:
Hi,

[...]

PR #684, opened 2023-08-24
Isothermal domain bc

baperry2:
This PR adds a robust isothermal wall capability. It addresses a need that has been brought up in various discussions/issues in the past (#263 #265 #502).

Isothermal walls are handled by separately computing the heat flux through walls after computing other diffusive fluxes (similar to how we handle EB isothermal wall fluxes). This avoids any issues with setting ghost cells that may affect the hydro calculations at the walls and also allows the values to be set directly on the face.

We see good agreement with and 2nd order convergence to the analytical solution for thermal diffusion from a wall into a semi-infinite medium `Tnorm = erf(x/sqrt(4*alpha*t))` without hydro. Note the test includes diffusion from all walls in the domain but the data is taken early enough that the effects of various walls haven't started interacting with each other yet. 

figure_dir0

figure_dir1

figure_convergence

Draft until documentation is added.


hsitaram (approved):
looking great! thank you Bruce

Docs/sphinx/BoundaryConditions.rst
```diff
@@ -18,7 +18,7 @@ PeleC manages boundary conditions in a form consistent with many AMReX codes. Gh
...
...Hard`` in the inputs will be tagged as ``EXT_DIR`` in ``pc_hypfill``.  Users will then fill the boundary values, by calling the helper function, ``bcnormal``. The ``bcnormal`` function fills an exterior (ghost) cell based on the value of the outermost interior cell. Its arguments include a problem-specific data structure, the location, direction, and orientation of the boundary being filled, and potentially fluctuating turbulent velocities from the `TurbInflow <https://amrex-combustion.github.io/PelePhysics/Utility.html#turbulent-inflows>`_ utility in PelePhysics. This gives the user flexibility to specify a variety of boundary conditions, including faces that contain both walls and inflow regions. Note that the external state ``s_ext`` is prepopulated with the same values as are used for the ``NoSlipWall`` condition, so the default if the ``bcnormal`` function does nothing is to specify a ``NoSlipWall``. Note that, to ensure conservation, when Gounov schemes are used the order of accuracy is reduced at boundaries specified using ``bcnormal``; PLM is used and the predictor step is omitted when computing fluxes through these boundaries. This does not affect any other boundary types.
```
marchdf: Would it be possible to highlight the "notes". Basically turning the last 2 sentences into `note` blocks.

Docs/sphinx/verification/verification.rst
```diff
@@ -247,3 +247,37 @@ constant Smagorinsky Large Eddy Simulation model.
 
 .. image:: /verification/les/p_error.png
    :width: 300pt
+
+Conservation and Isothermal Boundaries
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A simple test cass labeled ``MassCons`` is used to verify mass and energy conservation for all major numerical schemes used in PeleC (MOL without slopes, MOL with slopes, Godunov PLM, Godunov PPM). A python script is also run for all applicable test cases without inflows, outflows, or forcing, to ensure conservation during regression testing to ensure the conservation is maintained. The results shown below demonstrate the machine-precision level conservation of mass and energy for all numerical schemes. The case is arbitrary flow in a box with different types of boundaries (``SlipWall``, ``NoSlipWall``, ``UserBC``, including isothermal options). Note energy is conserved for the isothermal case without hydro turned on because of symmetry in the boundary conditions, but is not conserved for the case with hydro because includion of hydrodynamic effects breaks the symmetry.
```
marchdf: typo: "includion" instead of "inclusion"

Source/Diffusion.cpp
```diff
@@ -239,6 +239,61 @@ PeleC::getMOLSrcTerm(
...
       }
 
+      if (do_isothermal_walls) {
+        // Compute extensive diffusion flux at domain boundaries
+        BL_PROFILE("PeleC::isothermal_wall_fluxes()");
+        for (int dir = 0; dir < AMREX_SPACEDIM; dir++) {
+          if (
+            (typ == amrex::FabType::singlevalued) ||
+            (typ == amrex::FabType::regular)) {
+            if (domlo_isothermal_temp[dir] > 0.0) {
```
marchdf: So if this is > 0, it will trigger the isothermal wall. But in the documentation, you say it will trigger on negative values. I don't see anyone setting 0K walls so maybe just tweak the docs a bit?

Source/Diffusion.cpp
```diff
@@ -239,6 +239,61 @@ PeleC::getMOLSrcTerm(
...
+                    pc_set_wall_temperature(
+                      i, j, k, dir, normal, domlo_temp, geomdata, *lprobparm,
+                      qar, temp_arr);
+                    pc_isothermal_wall_fluxes(
+                      i, j, k, dir, normal, qar, temp_arr, flag_arr,
+                      area_arr[dir], flx[dir], dxinv, ltransparm);
+                  });
+              }
+            }
+            if (domhi_isothermal_temp[dir] > 0.0) {
```
marchdf: Is the only difference btw these 2 code blocks the value of `normal`? Maybe do an OR check `domhi_isothermal_temp[dir] || domlo_isothermal_temp[dir]` and then flip the normal based on that? Trying to reduce the duplicate code...

marchdf: You could also do a loop on Orientation and index into a vector containing the domlo/domhi stuff

Docs/sphinx/BoundaryConditions.rst
baperry2: Will do.

Source/Diffusion.cpp
baperry2: Will tweak the Docs.

Source/Diffusion.cpp
baperry2: Done as a loop

marchdf (approved):

PR #692, opened 2023-09-18
Update amrex

marchdf:
This should fix the FPE we are seeing in the Converging Nozzle test case.


jrood-nrel (approved):

jrood-nrel:
https://github.com/AMReX-Combustion/PeleC/pull/693 Will do this as well as hopefully fix HIP, so I will close this.

PR #698, opened 2023-09-28
Change default to not use slopes in SRD for robustness

baperry2:
Slopes are limited individually in the AMReX implementation of SRD. That's causes problems because the limiting can be inconsistent between energy and density and cause negative temperatures. For several cases, we've found it necessary to zero the slopes by setting `pelec.eb_srd_max_order = 0` to avoid this problem. This PR makes that the default option to provide a more robust experience for users. The existing behavior is still available to advanced users by setting `pelec.eb_srd_max_order = 2`. 

Comparison of the two approaches in the diverging section of a simple 2D nonreacting inviscid nozzle case:

`pelec.eb_srd_max_order = 0`
Screen Shot 2023-09-28 at 2 54 42 PM

`pelec.eb_srd_max_order = 2` (temperature goes negative and NaNs come on the next timestep)
Screen Shot 2023-09-28 at 2 54 28 PM

This should be expected to cause diffs in the nightly tests for most EB cases.


marchdf (approved):

baperry2:
Comparing results before and after this change, it appears to be making many of the regtests better behaved by the eyeball metric, or having negligible impact in some cases. Here are some examples:

EB-BluffBody Before:
Screen Shot 2023-09-29 at 1 54 57 PM

EB-BluffBody After:
Screen Shot 2023-09-29 at 1 54 47 PM

Reduced Temperature Overshoot in EB-C4:
Screen Shot 2023-09-29 at 2 19 43 PM

PR #702, opened 2023-10-06
Reproducibility for EB covered state

baperry2:
The state in cells covered by the EB is currently is set based on the value of the state in a somewhat arbitrarily selected fluid cell. Right now, the lexicographically first fluid cell in the first FAB containing fluid cells on the highest rank processor with a fluid cell is selected to populate the covered state. 

The identity of this cell depends on the domain decomposition and therefore the covered state may change for a particular case when the number of processors or `amr.max_grid_size` is changed. See below, for EB-C8 in 2D with 1 and 4 processors.

With this PR, the lexicographically first fluid cell in the entire domain is chosen, so it is no longer depends on domain decomposition. As a result, both cases below will now be equivalent and match the `1 proc` version. The code to do this is not very pretty, but I wasn't sure of a better way.

This PR should have no effect on the actual solutions, but may lead to some diffs due to covered cells changing.

`1 proc, current`
Screen Shot 2023-10-06 at 3 51 37 PM

`4 proc, current`
Screen Shot 2023-10-06 at 3 51 09 PM

### DIFF
--- a/Source/InitEB.cpp
+++ b/Source/InitEB.cpp
@@ -239,22 +239,32 @@ PeleC::define_body_state()
 
   // Scan over data and find a point in the fluid to use to
   // set computable values in cells outside the domain
+  // We look for the lexicographically first valid point
   if (!body_state_set) {
-    bool foundPt = false;
     const amrex::MultiFab& S = get_new_data(State_Type);
     auto const& fact =
       dynamic_cast<amrex::EBFArrayBoxFactory const&>(Factory());
     auto const& flags = fact.getMultiEBCellFlagFab();
+    amrex::Vector<int> h_minvec(AMREX_SPACEDIM, 0);
+    AMREX_D_TERM(h_minvec[0] = geom.Domain().bigEnd(0);
+                 , h_minvec[1] = geom.Domain().bigEnd(1);
+                 , h_minvec[2] = geom.Domain().bigEnd(2);)
 
-    for (amrex::MFIter mfi(S, false); mfi.isValid() && !foundPt; ++mfi) {
+    amrex::Gpu::DeviceVector<amrex::Real> local_body_state(NVAR, -1);
+    amrex::Gpu::DeviceVector<int> d_minvec(AMREX_SPACEDIM, 0);
+    amrex::Gpu::copy(
+      amrex::Gpu::hostToDevice, h_minvec.begin(), h_minvec.end(),
+      d_minvec.begin());
+
+    for (amrex::MFIter mfi(S, false); mfi.isValid(); ++mfi) {
       const amrex::Box vbox = mfi.validbox();
       auto const& farr = S.const_array(mfi);
       auto const& flag_arr = flags.const_array(mfi);
 
       const auto lo = amrex::lbound(vbox);
       const auto hi = amrex::ubound(vbox);
-      amrex::Gpu::DeviceVector<amrex::Real> local_body_state(NVAR, -1);
       amrex::Real* p_body_state = local_body_state.begin();
+      int* p_minvec = d_minvec.begin();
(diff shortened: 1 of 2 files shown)


marchdf:
Ah this is great. Thank you for looking at this. Long been on my todo list... I think there's a better way though. Let's chat.

marchdf (approved):

Source/InitEB.cpp
```diff
@@ -239,59 +239,37 @@ PeleC::define_body_state()
...
-              if (flag_arr(iv).isRegular()) {
-                found = true;
-                for (int n = 0; n < NVAR; ++n) {
-                  p_body_state[n] = farr(iv, n);
-                }
-              }
-            }
-          }
-        }
+    amrex::IntVect bx_size = geom.Domain().size();
```
marchdf: `const auto &` and maybe call it `dom_size` ?

Source/InitEB.cpp
```diff
@@ -239,59 +239,37 @@ PeleC::define_body_state()
...
+    long max_idx = geom.Domain().numPts();
```
marchdf: `const auto` or `const long`

Source/InitEB.cpp
```diff
@@ -239,59 +239,37 @@ PeleC::define_body_state()
...
+
+    // function minimized at the uncovered cell
+    amrex::ParallelFor(
+      minIdxTmp, [=] AMREX_GPU_DEVICE(int nbx, int i, int j, int k) noexcept {
+        tmp_arrays[nbx](i, j, k) =
+          flag_arrays[nbx](i, j, k).isRegular()
+            ? i * AMREX_D_TERM(1, *bx_size[1], *bx_size[2]) +
```
marchdf: you can also do something like this `const auto icell = tbox.index(iv);`. That gets you a 1D index of the `iv` in a box, which you could execute on the domain box.

Source/InitEB.cpp
```diff
@@ -239,59 +239,37 @@ PeleC::define_body_state()
...
-    found[amrex::ParallelDescriptor::MyProc()] = (int)foundPt;
-    amrex::ParallelDescriptor::ReduceIntSum(
-      found.data(), static_cast<int>(found.size()));
-    int body_rank = -1;
-    for (int i = 0; i < found.size(); ++i) {
-      if (found[i] == 1) {
-        body_rank = i;
-      }
+    // select the data at the first uncovered cell
+    amrex::IntVect idx_min = minIdxTmp.minIndex(0);
```
marchdf: const

Source/InitEB.cpp
```diff
@@ -239,59 +239,37 @@ PeleC::define_body_state()
...
+    amrex::Box tgt_box{idx_min, idx_min};
```
marchdf: const

marchdf (changes requested):

Source/InitEB.cpp
```diff
@@ -239,59 +239,37 @@ PeleC::define_body_state()
...
+    long max_idx = geom.Domain().numPts();
+
+    // function minimized at the uncovered cell
+    amrex::ParallelFor(
+      minIdxTmp, [=] AMREX_GPU_DEVICE(int nbx, int i, int j, int k) noexcept {
+        tmp_arrays[nbx](i, j, k) =
+          flag_arrays[nbx](i, j, k).isRegular()
+            ? i * AMREX_D_TERM(1, *bx_size[1], *bx_size[2]) +
+                j * AMREX_D_TERM(1, *1, *bx_size[2]) + k
+            : max_idx;
```
marchdf: `static_cast<amrex::Real>(max_idx)`

marchdf (approved):

PR #704, opened 2023-10-18
Some improvements and testing for the EB covered state

### LINKED ISSUE #703: Ensure unnecessary operations are not carried out in covered regions
PR https://github.com/AMReX-Combustion/PeleC/pull/702 ensures consistency in setting the state in covered cells. Presently, valid data state is required in covered cells because some operations, such as `bcnormal` to populate ghost cells at domain boundaries, are called for covered cells, even though that's not strictly necessary and shouldn't affect the solution. We should remove these unneeded operations (perhaps by temporarily setting the covered state to something invalid, like all zeros, to chase down any such instances).

baperry2:
This closes https://github.com/AMReX-Combustion/PeleC/issues/703, with the goal of ensuring that covered cells never affect fluid cell values:

- A change in behavior for all simulations - if a NaN is detected in the state after a timestep, the simulation now stops
- For all EB stencils, if a value in the stencil is zero, that element of the stencil is ignored rather than multiplying it by the value in the cell and adding it into the gradient being computed. This should not effect the solution, but is done so that if the value is NaN in a covered cell, it does not propagate into the domain.
- A new function is added to abort if the stencil for gradients at the boundary tries to access covered cells (can happen for the quadratic option on coarse grids `ebd.boundary_grad_stencil_type = 0`)
- the function `zero_in_body` has been deleted because it was not used anywhere
- A new option to set the body state to 0 for all variables for debugging (to ensure this does not impact fluid cells)
- The eb-c3 test is modified to set the body state to 0. This is done to prevent any future changes that would cause data from covered cells to be accessed during the fluid cell updates. Had to make a few changes to `Tests/CMakeLists.txt` because this test can't run with AMReX FPE trapping.


hsitaram (approved):
looks good! thank you, Bruce.

baperry2:
The end of timestep NaN check works at least on AMD and Nvidia GPUs (Frontier and Eagle). 

On GPU, the abort for invalid stencils only works in debug mode or with USE_ASSERTION=TRUE.

Did not notice a change in performance.

Tests/CMakeLists.txt
```diff
@@ -55,11 +55,11 @@ macro(setup_test)
     file(COPY ${CURRENT_TEST_SOURCE_DIR}/${TEST_NAME}.inp DESTINATION "${CURRENT_TEST_BINARY_DIR}/")
     file(COPY ${TEST_FILES} DESTINATION "${CURRENT_TEST_BINARY_DIR}/")
     # Set some default runtime options for all tests
-    set(RUNTIME_OPTIONS "amr.plot_file=plt amr.checkpoint_files_output=0 amr.plot_files_output=1 amrex.the_arena_is_managed=0")
+    set(RUNTIME_OPTIONS_AlL "amr.plot_file=plt amr.checkpoint_files_output=0 amr.plot_files_output=1 amrex.the_arena_is_managed=0")
```
jrood-nrel: There is a typo here.

Tests/CMakeLists.txt
```diff
@@ -113,6 +113,14 @@ function(add_test_rf TEST_NAME TEST_EXE_DIR)
     set_tests_properties(${TEST_NAME} PROPERTIES WILL_FAIL TRUE)
 endfunction(add_test_rf)
 
+# Regression tests no FPE trapping ever
```
jrood-nrel, line 116: This test won't have these options `amr.plot_file=plt amr.checkpoint_files_output=0 amr.plot_files_output=1 amrex.the_arena_is_managed=0`, which I think we would want to enforce.

jrood-nrel: Actually I'm confused. It just has the same options as the other tests?

jrood-nrel: You could leave the `RUNTIME_OPTIONS` as is and just add `amrex.signal_handling=0` to this test category.

Docs/sphinx/geometry/EB.rst
```diff
@@ -162,6 +162,12 @@ Applying boundary and face stencils
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 When processing geometry cells, the cached datastructures can be applied efficiently, for example, to interpolate fluxes from face centers to face centroids in cut cells.
+Two stencil types are available for computing the gradients at the EB for diffusive fluxes, and they are selected as follows:
+
+* ``ebd.boundary_grad_stencil_type = 0``: Quadratic stencil (default). On poorly resolved geometries, this stencil may reach into covered cells, in which case the simulation will
+  fail with a warning. See `Johansen and Collela <https://doi.org/10.1006/jcph.1998.5965>`_ for further details.
+
+* ``end.boundary_grad_stencil_type = 1``: Least-squares stencil. See `Anderson and Bonhaus <https://doi.org/10.1016/0045-7930(94)90023-X>`_ for further details.
```
marchdf: typo: should be `ebd`

Source/PeleC.cpp
```diff
@@ -1407,6 +1408,14 @@ PeleC::okToContinue()
                    << std::endl;
   }
 
+  if (
+    get_new_data(State_Type).contains_nan() or
```
marchdf: can you use `||` instead of `or`?

Source/EB.cpp
```diff
@@ -389,6 +389,44 @@ pc_fill_bndry_grad_stencil_ls(
...
+      for (int ii = 0; ii < 3; ii++) {
+        for (int jj = 0; jj < 3; jj++) {
+#if AMREX_SPACEDIM > 2
+          for (int kk = 0; kk < 3; kk++) {
+#endif
+            const amrex::IntVect ivp = amrex::IntVect(AMREX_D_DECL(
+              sten[L].iv_base[0] + ii, sten[L].iv_base[1] + jj,
+              sten[L].iv_base[2] + kk));
+            if (
+              flags(ivp).isCovered() and
```
marchdf: `&&` instead of `and`

marchdf (approved):

jrood-nrel (approved):

PR #707, opened 2023-10-20
Fix derive for state variables with ghost cells

### LINKED ISSUE #706: AMR refinement based on adjacent_difference_greater is broken
Test case:

Run any existing PMF case in PeleC, remove existing refinement criteria, and add, for example, the following refinement criteria:
`tagging.refinement_indicators = gtemp`
`tagging.gtemp.adjacent_difference_greater = 10`
`tagging.gtemp.field_name = Temp`
`tagging.gtemp.max_level = 1`

Result:

Running with AMR max level set to 0, the simulation runs fine. Setting the max level to 1 breaks the simulation with a segfault. Turning on debug mode shows that a 4 dimensional array (state?) is being referenced out of bounds (-1:16, -1:16, -1:16, 0:0).

baperry2:
When deriving a state variable with ghost cells, the derive function in AmrLevel does a FillPatch for the individual state variable. However, in PeleC, the `PCHypFillExtDir` function can only operate on the full state, not individual variables. Therefore, instead of using the AmrLevel derive function in this case, we should manually FillPatch the whole state and then take the component that we need. 

Resolves https://github.com/AMReX-Combustion/PeleC/issues/706 (the only time this problem comes up right now is when using the `adjacent_difference_greater` refinement criterion with a state variable, which requires a ghost cell). Also cleans up the derive for LES variables, adds error checking to `PCHypFillExtDir` to ensure it is only called for the full state, and adds `adjacent_difference_greater` to a test case.

### DIFF
--- a/Source/BCfill.cpp
+++ b/Source/BCfill.cpp
@@ -30,8 +30,7 @@ struct PCHypFillExtDir
     const int /*orig_comp*/) const
   {
     // We need the whole state to apply physical BCs
-    AMREX_ASSERT(dcomp == 0 and numcomp == NVAR);
-    amrex::ignore_unused(dcomp, numcomp);
+    AMREX_ALWAYS_ASSERT(dcomp == 0 && numcomp == NVAR);
 
     const int* domlo = geom.Domain().loVect();
     const int* domhi = geom.Domain().hiVect();
--- a/Source/BCfill.cpp
+++ b/Source/BCfill.cpp
@@ -21,14 +21,18 @@ struct PCHypFillExtDir
   void operator()(
     const amrex::IntVect& iv,
     amrex::Array4<amrex::Real> const& dest,
-    const int /*dcomp*/,
-    const int /*numcomp*/,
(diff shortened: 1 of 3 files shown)


Source/BCfill.cpp
```diff
@@ -21,14 +21,18 @@ struct PCHypFillExtDir
...
+    const int dcomp,
+    const int numcomp,
     amrex::GeometryData const& geom,
     const amrex::Real time,
     const amrex::BCRec* bcr,
     const int /*bcomp*/,
     const int /*orig_comp*/) const
   {
+    // We need the whole state to apply physical BCs
+    AMREX_ASSERT(dcomp == 0 and numcomp == NVAR);
```
marchdf: you could switch this to an `AMREX_ALWAYS_ASSERT` and then you don't need the ignore_unused. Might be good to always assert this one. 

Also, can you use `&&` and not `and` ?

marchdf (approved):

PR #710, opened 2023-11-02
remove dependence of Filter on Utilities

baperry2:
This basically makes the Filter class independent of the rest of PeleC, so that it can be easily pulled in and used by external tools (PeleAnalysis).

### DIFF
--- a/Source/Filter.cpp
+++ b/Source/Filter.cpp
@@ -473,8 +473,9 @@ Filter::apply_filter(
   const int /*ncomp*/)
 {
   const auto q = in.const_array();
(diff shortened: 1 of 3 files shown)


marchdf (approved):

Source/Filter.cpp
```diff
@@ -473,8 +473,9 @@ Filter::apply_filter(
   const int /*ncomp*/)
 {
   const auto q = in.const_array();
+  out.setVal(0.0, box, nstart, ncnt);
```
marchdf: I think you need one of those "runon" things. E.g. like in Diffusion

PR #725, opened 2023-12-13
abort on unused parmparse variables in input file in tests

baperry2:
So excess/deprecated inputs don't build up and cause confusion.


baperry2:
@marchdf @jrood-nrel how do we feel about this? I haven't gone through all the input files yet, but the ones changed here are representative. It catches some deprecated and/or irrelevant inputs, but mostly just catches `amr.ref_ratio` and amr.regrid_int` for cases without AMR.

jrood-nrel:
Love it.

marchdf:
I like it. I wish we could ignore the ref_ratio and regrid_int when max_level = 0. It's less stuff to change when we change the max level... but maybe that's too much to ask?

baperry2:
We could put dummy queries for these just so they always get touched, but I'm not a big fan of that. IMO not a big deal just to uncomment those lines when adding levels.

jrood-nrel (approved):

marchdf (approved):

marchdf:
yeah probably fine.

baperry2:
@marchdf based on what this uncovered, it looks like the forcing for HIT case wasn't actually active, which I think I fixed. Also, the "moving plane" EB geometry and corresponding EB-Plane test seem to not really do anything, so I deleted them. Can you double check these changes?

jrood-nrel (approved):

marchdf (approved):

marchdf:
nice work!

PR #730, opened 2023-12-19
Remove prob parm passing in spray

marchdf:


marchdf:
@jrood-nrel is going to take this PR from here to fix the cmake stuff. But we can merge https://github.com/AMReX-Combustion/PelePhysics/pull/462 in the mean time.

Exec/RegTests/Spray-Conv/SprayParticlesInitInsert.cpp
```diff
@@ -9,18 +9,13 @@ SprayParticleContainer::injectParticles( // NOLINT
...
+  int /*finest_level*/)
 {
   return false;
 }
 
 void
-SprayParticleContainer::InitSprayParticles(
-  const bool init_parts,
-  ProbParmHost const& /*prob_parm*/,
-  ProbParmDevice const& /*prob_parm_d*/)
```
baperry2, line 23: As discussed [here](https://github.com/AMReX-Combustion/PeleMP/pull/17/files#r1424431642), we should add a pp query in this function for "parcel_size", and add that as the last (optional) argument to unifromSprayInit, which would allow this test to recover its behavior its behavior from before we started all this reorganization.

marchdf: uh. Can you make that happen ;) ? I am lost

baperry2: done

marchdf: thank you!

jrood-nrel (approved):

PR #731, opened 2023-12-20
MD's xmas present: EB Godunov

marchdf:
Seems to work ok for a shock hitting a cylinder

Screenshot 2023-12-20 at 4 35 08 PM


jrood-nrel (approved):

Source/Hydro.H
```diff
@@ -73,42 +479,504 @@ pc_divu(
...
+{
+  amrex::Real pdivu;
+
+#if (AMREX_SPACEDIM == 2)
+  if (flag(i, j, k).isRegular()) {
+    pdivu = 0.5 * ((q1(i + 1, j, k, GDPRES) + q1(i, j, k, GDPRES)) *
+                     (q1(i + 1, j, k, GDU) - q1(i, j, k, GDU)) * dxinv[0] +
+                   (q2(i, j + 1, k, GDPRES) + q2(i, j, k, GDPRES)) *
+                     (q2(i, j + 1, k, GDV) - q2(i, j, k, GDV)) * dxinv[1]);
+    divu(i, j, k, UEINT) += pdivu;
```
marchdf: @baperry2 I think this (and the others in this function) should be `-= pdivu;` since that would follow what we do for the non-eb path. But flipping the sign doesn't change anything and I even multiplied by 1e12 and nothing happens. Maybe we just don't use this...

drummerdoc: Going on very old memories, but might the "artificial viscosity" be zero in the cases you're testing with?

marchdf: hmmm this comment sent me into a tailspin. I've got bug fixes out to CAMR as well.  I think the present is going to be late... but it's going to happen!

drummerdoc: I hear some people give gifts for New Year's.

Source/PeleC.cpp
```diff
@@ -403,9 +403,9 @@ PeleC::read_params()
   soot_model.readSootParams();
 #endif
 
-  if ((!do_mol) && eb_in_domain) {
-    amrex::Abort("Must do_mol = 1 when using EB\n");
-  }
+  // if ((!do_mol) && eb_in_domain) {
+  //   amrex::Abort("Must do_mol = 1 when using EB\n");
+  // }
```
baperry2: Delete rather than comment to avoid later confusion.

Source/Hydro.cpp
```diff
@@ -238,7 +238,8 @@ PeleC::construct_hydro_source(
             p_rrflag_as_crse->array(), as_fine, dm_as_fine.array(),
             level_mask.const_array(mfi), dt, ppm_type, plm_iorder,
             use_flattening, difmag, bcs_d.data(), redistribution_type,
-            eb_weights_type, cflLoc);
+            eb_weights_type, eb_srd_max_order, eb_clean_massfrac,
```
baperry2, line 243: pc_umdrv_eb needs a BL_PROFILE statement

Source/Godunov.cpp
```diff
@@ -453,6 +475,565 @@ pc_umeth_3D(
...
+
+  amrex::FArrayBox qxzm(xmbx, QVAR, amrex::The_Async_Arena());
+  amrex::FArrayBox qxzp(xmbx, QVAR, amrex::The_Async_Arena());
+  auto const& qmxz = qxzm.array();
+  auto const& qpxz = qxzp.array();
+
+  //
+  amrex::Box xybx(bxg2);
+  xybx.grow(1, -1);
+  // amrex::Print() << "TRANSDO: Y FLUXES CHANGING X " << xybx << std::endl;
```
baperry2: Delete commented debugging (there are lots of these)

Source/Godunov.cpp
```diff
@@ -453,6 +475,565 @@ pc_umeth_3D(
...
+  // amrex::Print() << "DOING CMPFLX XZ " << xzcmpbx << std::endl;
+  amrex::ParallelFor(
+    xzcmpbx, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
+      // X|Z
+      if (apx(i, j, k) > 0.) {
+        pc_cmpflx(
+          i, j, k, bclx, bchx, dlx, dhx, qmxz, qpxz, flxz, qxz, qaux, cdir);
+      }
+    });
+  qxym.clear();
```
baperry2: Are all these clears necessary? We don't have them in the non-EB version

Source/Godunov.cpp
```diff
@@ -453,6 +475,565 @@ pc_umeth_3D(
...
+  // do tangential interpolation when taking divergence later
+  amrex::Box zfbx(surroundingNodes(bx_to_fill, 2));
+  zfbx.grow(0, 1);
+  zfbx.grow(1, 1);
+  amrex::ParallelFor(zfbx, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
+    if (apz(i, j, k) > 0.) {
+      pc_cmpflx(i, j, k, bclz, bchz, dlz, dhz, qm, qp, flx3, q3, qaux, cdir);
+    }
+  });
+  qmfab.clear();
```
baperry2: the `pc_low_order_boundary` stuff from the non-EB version needs to be added or we won't conserve mass for bcnormal boundaries

Source/Godunov.cpp
```diff
@@ -654,4 +1247,200 @@ pc_umeth_2D(
...
+
+  // Final Riemann problem Y
+  amrex::ParallelFor(yfxbx, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
+    if (apy(i, j, k) > 0.) {
+      pc_cmpflx(
+        i, j, k, bcly, bchy, dly, dhy, qmarr, qparr, flx2, q2, qaux, cdir);
+    }
+  });
+}
+
```
baperry2, line 1451: Again, we need the `pc_low_order_boundary` stuff to ensure mass conservation

Source/Hydro.H
```diff
@@ -73,42 +475,506 @@ pc_divu(
...
-    0.25 *
-    (q(i, j, k, QW) - q(i, j, k - 1, QW) + q(i - 1, j, k, QW) -
-     q(i - 1, j, k - 1, QW) + q(i, j - 1, k, QW) - q(i, j - 1, k - 1, QW) +
-     q(i - 1, j - 1, k, QW) - q(i - 1, j - 1, k - 1, QW)) /
-    dz;
+  const amrex::Real ux = (q(i, j, k, QU) - q(i - 1, j, k, QU)) / dx;
+#elif (AMREX_SPACEDIM == 2)
+  amrex::Real ux, vy;
+
+  if (j == domlo_y && bclo_y == NoSlipWall) {
```
baperry2: Here and below, I'm concerned about all the checks on the boundary conditions. They were added in CAMR so that artificial viscosity in simulations with No-Slip walls would not affect the velocity at the walls. While that is a valid choice to make, I think it's also fine to say that artificial viscosity should act like regular viscosity, and a no-slip wall should therefore be able to introduce shear through artificial viscosity. 

My concern is that here the check is only done on NoSlipWall, but not bcnormal that may actually be enforcing a NoSlipWall. I'd prefer just to leave this function as it was before.

Source/Hydro.H
baperry2: Looks like you resolved this in CAMR, so I'm fine with this if it macthes that solution.

baperry2 (changes requested):
See other comments for various requested changes, the biggest of which is that the boundary corrections for bcnormal boundaries need to be applied or we will not conserve mass. I will modify the MassCons regtests so that they include EB and capture this issue.

A few other questions comments:
- Should this be expected to play nicely with the re-re stuff - can we have coarse/fine boundaries on the EB with EB-Godunov? If so, can the test case include that?
- I did some tests where the EB intersects a domain boundary. Similar to EB-MOL, depending on how the EB intersects the boundary, there can be mass leakage issues or the simulation can crash. Things seem to be ok if the EB is normal to the domain boundary.
- It seems weird to add such a big capability and not add any documentation, although that's mostly because the current documentation doesn't say Godunov isn't supported with EB. Still, I think it would be good to add a paragraph somewhere that indicates Godunov is preferred numerically but MOL is more extensively tested with EB (and both may have issues when EB intersects domain boundary at an angle).
- The EB-Godunov capability being tested only in the Sod regtest is weird. Can we add this to one of the EB-* cases as well (just to make it easier for users to find the capability)

Source/Hydro.cpp
marchdf: good catch, done.

Source/Hydro.H
marchdf: Reverted to previous.

Source/PeleC.cpp
marchdf: done

Source/Godunov.cpp
marchdf: removed

marchdf:
Some answers:
- yes it will play nice with the re-re stuff. I will add a test
- thank you for adding those mass conservation tests
- I will add documentation that you suggest
- yes, good point

TODO based on comments:
- [x] add re-re test
- [x] add documentation
- [x] add more test like some of the EB-* ones
- [x] include `pc_low_order_boundary` stuff

Source/Godunov.cpp
marchdf: done

Source/Godunov.cpp
marchdf: done

baperry2 (approved):

PR #743, opened 2024-01-12
Better Problem Specific Function capability

baperry2:
Consolidate problem-specific functions into a single class, which makes adding more problem-specific capability easier. This is a breaking change, only requires a one-line fix for case that don't use the existing problem-specific capability, and makes it so future additions of problem-specific capability aren't breaking. 

Also adds some improvements for passive scalar advection.

Draft right now, will need to go through and make a change to all cases once the methodology is finalized.

@marchdf - There are two potential approaches that I'm considering. The first is to make all problem-specific functions (whether they run on CPU or GPU) into static functions of a single class. The second is to make all CPU problem-specific functions be members of that class, but have the problem-specific GPU kernels be stand alone functions that use templating based on that class. See `Source/ProblemSpecificFunctions.H` and `MassCons/prob.H` for examples (`set_problem_tags` is the former, `set_isothermal_wall_temperature` is the latter). Let me know which approach you prefer.


baperry2:
This should be ready to go now. The first approach described above was chosen.

Source/Diffusion.cpp
```diff
@@ -175,13 +176,33 @@ PeleC::getMOLSrcTerm(
...
+            amrex::Real muloc, xiloc, lamloc;
+            amrex::Real Ddiag[NUM_SPECIES], Y[NUM_SPECIES] = {0.0};
+            amrex::Real* chi_mix = nullptr;
+            amrex::Real T = qar_Tin(i, j, k);
+            amrex::Real rho = qar_rhoin(i, j, k);
+            for (int n = 0; n < NUM_SPECIES; ++n) {
+              Y[n] = qar_yin(i, j, k, n);
+            }
+
+            amrex::RealVect x = pc_cmp_loc({AMREX_D_DECL(i, j, k)}, geomdata);
```
marchdf: You could `const` all these I think?

Source/Diffusion.cpp
```diff
@@ -175,13 +176,33 @@ PeleC::getMOLSrcTerm(
...
+            pc_transcoeff(
+              get_xi, get_mu, get_lam, get_Ddiag, get_chi, T, rho, Y, Ddiag,
+              chi_mix, muloc, xiloc, lamloc, ltransparm, *lprobparm, x);
```
marchdf, line 198: ah I get the trickiness you had to figure out... I guess the only bummer is the extra stuff that needs to be initialized/used here. Duplicate stuff from pelephysics. But there's probably no other good way.

Source/Utilities.H
```diff
@@ -3,10 +3,23 @@
...
 #include "Constants.H"
 #include "IndexDefines.H"
 #include "PelePhysics.H"
 
+AMREX_GPU_HOST_DEVICE
+AMREX_FORCE_INLINE
+amrex::RealVect
+pc_cmp_loc(const amrex::IntVect& idx, const amrex::GeometryData& geomdata)
+{
+  amrex::RealVect x;
```
marchdf: You could make this a one-liner.
```
return {AMREX_D_DECL(geomdata.ProbLo(0) + (idx[0] + 0.5) * geomdata.CellSize(0), geomdata.ProbLo(1) + (idx[1] + 0.5) * geomdata.CellSize(1), geomdata.ProbLo(2) + (idx[2] + 0.5) * geomdata.CellSize(2))}
```

marchdf (approved):

marchdf:
Great work! Well done. I had some minor comments.

PR #774, opened 2024-03-28
Implement the state prescription at EB for the Riemann solver

marchdf:
TODO:
- [x] document this and warn users about it (it's beta, only right for hydro fluxes)
- [x] add an abort if using godunov
- [x] add a user defined flag "do_problem_eb_fill" that gets passed to `pc_compute_hyp_mol_flux_eb`
- [x] move `ebfill` into the problem specific function stuff
- [x] test on eb-inflow case


Source/MOL.cpp
```diff
@@ -231,9 +247,137 @@ pc_compute_hyp_mol_flux_eb(
...
+      qtempr[R_UT2] = qtempl[R_UT2];
+      qtempr[R_P] = qtempl[R_P];
+      amrex::Real spr[NUM_SPECIES] = {0.0};
+      for (int n = 0; n < NUM_SPECIES; n++) {
+        spr[n] = spl[n];
+      }
+      amrex::Real rhoe_r = rhoe_l;
+      amrex::Real gamc_r = gamc_l;
+
+      const bool do_ebfill = ProblemSpecificFunctions::ebfill(
```
baperry2: I don't like that we unnecessarily initialize all the arrays above even when do_ebfill ends up being false. How about instead of having do_ebfill be the return of the ebfill function, have it be a member of the PeleC class that gets parmparsed, so we can skip all the stuff above when false?

This code ends up being annoyingly similar to the non-EB function, but maybe just different enough that it's not possible to consolidate and reduce copy/pasted code.

marchdf: yeah... I can't remember the reason why I did some of these things. One thing I remember is that I wanted to know if a given region of the EB would be custom-filled or not. So instead of doing the test, then filling the arrays, I did all at once. There's probably a better way, let's chat about it next week.

baperry2 (approved):

PR #802, opened 2024-05-26
Additional SGS Models

JhonCordova:
The [WALE](https://www.researchgate.net/publication/227042556_Subgrid-Scale_Stress_Modelling_Based_on_the_Square_of_the_Velocity_Gradient_Tensor) and [Vreman](http://www.vremanresearch.nl/Vreman-PF2004-subgridmodel.pdf) models were integrated into the PeleC based on the rationale behind the original implementation of the Smagorinsky model available in this solver. Note that I have tried to encapsulate part of the code used in the 'getSmagorinskyLESTerm' method. Therefore, if there are any alternative suggestions, I am open to new ideas.


marchdf:
You can remove all the files in `refdata` most likely. We don't need extra copies of those.

Source/PeleC.cpp
```diff
@@ -2140,6 +2146,7 @@ PeleC::init_mms()
 #endif
     masa_init("mms", masa_solution_name.c_str());
     masa_set_param("Cs", PeleC::Cs);
+    masa_set_param("Cw", PeleC::Cw);
```
marchdf: This is MASA specific. You don't need this.

marchdf:
Thank you for doing this! We will take a look at it this week. If you want to get the CI to start passing, you will need to do some formatting using clang-format.

JhonCordova:
Thank you for the feedback. I will address the comments shortly.

baperry2:
Thanks for doing this, and sorry it's taken a while for us to review.

It looks like the `HIT_LES` case is just a copy of the `HIT` case, except the input file. Is that right? If so, you can just put a `hit-les.inp` input file in the `HIT` case to avoid duplication of all those files. If you do that, I'll add it to the testing suite to ensure that the new LES models do not break in the future.

Have you run some tests with the WALE/Vremen models, particularly in comparison to Smagorinsky? If so, posting plots of the results here would be helpful.

Source/LES.H
```diff
@@ -518,5 +518,591 @@ pc_dynamic_smagorinsky_sfs_term(
...
+  flx(iv, UMX) = flx(iv, UMX) * a(iv);
+  flx(iv, UMY) = flx(iv, UMY) * a(iv);
+  flx(iv, UMZ) = flx(iv, UMZ) * a(iv);
+  flx(iv, UEDEN) = flx(iv, UEDEN) * a(iv);
+}
+
+AMREX_GPU_DEVICE
+AMREX_FORCE_INLINE
+void
+get_vreman_sfs_stresses_xdir(
```
baperry2, line 838: Can these be simplified to have one function that takes `const int dir` as an input rather than having three separate functions to reduce duplicate code? I know the separate function for x/y/z is used elsewhere in LES, so perhaps its ok to leave this for now and clean it all up later.

Source/LES.cpp
```diff
@@ -255,27 +322,11 @@ PeleC::getSmagorinskyLESTerm(
...
       // Refluxing
-      if (do_reflux && reflux_factor != 0) {
-        amrex::FArrayBox dm_as_fine(
-          amrex::Box::TheUnitBox(), LESTerm.nComp(), amrex::The_Async_Arena());
-        update_flux_registers(
-          reflux_factor * dt, mfi, typ,
-          {AMREX_D_DECL(&flux_ec[0], &flux_ec[1], &flux_ec[2])}, dm_as_fine);
-      }
+      updateFluxRegistersLES(
+        reflux_factor, reflux_factor, LESTerm, dt, mfi, typ, flux_ec);
```
baperry2: It looks like the first argument here (and other calls to this function) should be `do_reflux`

JhonCordova:
> Thanks for doing this, and sorry it's taken a while for us to review.
> 
> It looks like the `HIT_LES` case is just a copy of the `HIT` case, except the input file. Is that right? If so, you can just put a `hit-les.inp` input file in the `HIT` case to avoid duplication of all those files. If you do that, I'll add it to the testing suite to ensure that the new LES models do not break in the future.
> 
> Have you run some tests with the WALE/Vremen models, particularly in comparison to Smagorinsky? If so, posting plots of the results here would be helpful.

Thanks for the feedback!

**1. HIT_LES Case Update:**
I've removed the HIT_LES case and included the hit-les.inp input file in the HIT case. This should streamline the process and avoid any unnecessary duplication of files. 

**2. SGS Models Comparison:**
I have conducted tests comparing the WALE and Vreman models to the Smagorinsky model. The results and detailed analysis are presented in my publication, which you can access [here](https://www.researchgate.net/publication/374585154_Assessment_of_Large_Eddy_Simulation_LES_Sub-grid_Scale_Models_Accounting_for_Compressible_Homogeneous_Isotropic_Turbulence). 

Let me know if there's anything else you need or if further testing is required.

Source/LES.cpp
JhonCordova: Thanks for the review. The modifications have been made.

Source/PeleC.H
```diff
@@ -380,6 +380,15 @@ public:
...
     amrex::Real flux_factor);
 
+  void updateFluxRegistersLES(
+    bool do_reflux,
+    amrex::Real reflux_factor,
+    const amrex::MultiFab& LESTerm,
+    amrex::Real dt,
+    const amrex::MFIter& mfi,
+    amrex::FabType typ,
+    amrex::FArrayBox flux_ec[AMREX_SPACEDIM]);
```
marchdf: I would change this to an `amrex::Array<>`. Also, is this function needed? Not sure I follow the use case.

marchdf:
@JhonCordova thanks for bearing with us as we wrangle this in. We really appreciate you contributing to the code and making it better. Good work!

baperry2 (approved):
@JhonCordova I have added the hit-les case with WALE to run in the CI tests. I also took the liberty of making some changes to resolve warnings, so I think all tests should pass now. Thanks for posting the link to your paper, that will be a good reference for anyone using the newly implemented models.

PR #805, opened 2024-05-28
Reaction mask implementation gpu

SreejithNREL:
Updated PeleC to include Chemical mask implementation in PelePhysics. The PR for that update has already been made.

### DIFF
--- a/Source/PeleC.H
+++ b/Source/PeleC.H
@@ -745,8 +745,7 @@ protected:
   static bool use_chem_mask;
   static amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> lo_chem_mask_coordinate;
   static amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> hi_chem_mask_coordinate;
-  static amrex::Gpu::DeviceVector<amrex::Real> d_lo_chem_mask_coordinate;
-  static amrex::Gpu::DeviceVector<amrex::Real> d_hi_chem_mask_coordinate;
+
 
   static bool use_typical_vals_chem;
   static bool use_typical_vals_chem_usr;
@@ -906,5 +905,6 @@ PeleC::getFluxReg(int lev)
 {
   return getLevel(lev).getFluxReg();
 }
+
 
 #endif
--- a/Source/PeleC.H
+++ b/Source/PeleC.H
@@ -746,7 +746,6 @@ protected:
   static amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> lo_chem_mask_coordinate;
   static amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> hi_chem_mask_coordinate;
 
-
   static bool use_typical_vals_chem;
   static bool use_typical_vals_chem_usr;
   static amrex::Real typical_rhoY_val_min;
@@ -905,6 +904,5 @@ PeleC::getFluxReg(int lev)
 {
   return getLevel(lev).getFluxReg();
 }
-
 
 #endif
--- a/Source/PeleC.H
+++ b/Source/PeleC.H
@@ -742,6 +742,10 @@ protected:
 
   static amrex::Vector<int> src_list;
 
+  static bool use_chem_mask;
+  static amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> lo_chem_mask_coordinate;
+  static amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> hi_chem_mask_coordinate;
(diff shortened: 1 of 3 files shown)


Source/PeleC.H
```diff
@@ -742,6 +742,10 @@ protected:
 
   static amrex::Vector<int> src_list;
 
+  static bool use_chem_mask;
```
marchdf, line 745: The best way to do this is actually a bit different. You would add the option(s) here and run the `Source/Params/mk_params.sh` script to make this all automatic. I will do the definition/declaration/parmparse for you.

Source/PeleC.cpp
```diff
@@ -250,6 +256,16 @@ PeleC::read_params()
   pp.query("use_typ_vals_chem", use_typical_vals_chem);
   pp.query("use_typ_vals_chem_usr", use_typical_vals_chem_usr);
 
+  pp.query("chem_mask", use_chem_mask);
+
+  // Reading chemistry mask box coordinates
+  if (use_chem_mask) {
+    for (int n = 0; n < AMREX_SPACEDIM; ++n) {
+      pp.get("lo_chemmask", lo_chem_mask_coordinate[n], n);
```
marchdf, line 264: let's be more explicit with this user facing string `lo_chemistry_mask` e.g.

Source/PeleC.H
marchdf: Can you also add some documentation about what these new input parameters mean for the user?

Source/React.cpp
```diff
@@ -126,6 +129,15 @@ PeleC::react_state(
 #pragma omp parallel if (amrex::Gpu::notInLaunchRegion())
 #endif
   {
+
+    amrex::RealBox Chem_Masked_Region(
```
marchdf, line 133: name this `masked_chemistry_box`, make it `const`.

Source/React.cpp
```diff
@@ -170,6 +182,20 @@ PeleC::react_state(
         auto const& mask = dummyMask.array(mfi);
         auto const& fc = fctCount.array(mfi);
 
+        if (use_chem_mask) {
+          amrex::ParallelFor(
+            bx, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
+              amrex::XDim3 point;
```
marchdf, line 188: can simplify: `const amrex::IntVect iv = {AMREX_D_DECL(plo[0] + (i + 0.5) * dx[0], etc)};` see other examples in the code.

marchdf: This should probably just happen once? No need to do it every step? if so, can we move to the init somewhere?

Source/React.cpp
```diff
@@ -170,6 +182,20 @@ PeleC::react_state(
         auto const& mask = dummyMask.array(mfi);
```
marchdf, line 182: while we are at it. Let's rename this "dummyMask` to something that means something.

Source/React.cpp
```diff
@@ -170,6 +182,20 @@ PeleC::react_state(
...
+        if (use_chem_mask) {
+          amrex::ParallelFor(
+            bx, [=] AMREX_GPU_DEVICE(int i, int j, int k) noexcept {
+              amrex::XDim3 point;
+              point.x = plo[0] + (i + 0.5) * dx[0];
+              point.y = plo[1] + (j + 0.5) * dx[1];
+              point.z = plo[2] + (k + 0.5) * dx[2];
+
+              if (Chem_Masked_Region.contains(point)) {
+                mask(i, j, k) = -1;
```
marchdf, line 194: You can then make this a ternary expression: `mask(i, j, k) = Chem_Masked_Region.contains(iv)? -1 : 0;`. It's also weird to me that `mask` is -1 when masked. Should it be 0? And then 1 for when unmasked. That would fit my mental definition of true/false and other uses of the term "mask" in amrex codes.

Source/React.cpp
```diff
@@ -209,7 +235,7 @@ PeleC::react_state(
         );
 
         amrex::Gpu::Device::streamSynchronize();
-
+        bool use_chem_mask_d = use_chem_mask;
```
marchdf, line 238: const this.

marchdf (changes requested):

Source/React.cpp
baperry2: I think @marchdf means `const amrex::RealVect = ...`. The AMREX_D_DECL is needed so the code also works in 2d. You can also use the pc_cmp_loc function which wraps this.

PR #808, opened 2024-06-03
fix interpolation in PMF cases

baperry2:
Fixes bug interpolating data from PMF files


marchdf (approved):

alexnovoselov:
This fixed the issues we were having with this case!

PR #819, opened 2024-06-25
use template to simplify a bunch of derive functions

baperry2:
No need to have a bunch of repeated code just to divide different variables by density, so let's get rid of some lines of code.


marchdf (approved):
I like it!

PR #827, opened 2024-07-16
EB-FlowOverCylinder Case

d-montgomery:
I added a case that flows Hagen–Poiseuille flow over a cylinder defined with embedded boundaries. The case utilizes the Godunov method instead of the MOL.


Exec/RegTests/EB-FlowPastCylinder/GNUmakefile
```diff
@@ -0,0 +1,38 @@
...
+TINY_PROFILE = FALSE
+COMM_PROFILE = FALSE
+TRACE_PROFILE = FALSE
+MEM_PROFILE = FALSE
+USE_GPROF = FALSE
+
+# Performance
+USE_MPI = TRUE
+USE_OMP = FALSE
+USE_CUDA = FALSE # invidia GPU
```
baperry2: remove comment

Exec/RegTests/EB-FlowPastCylinder/prob.H
```diff
@@ -0,0 +1,144 @@
...
+    amrex::Real rho = 0.0, u = 0.0, v = 0.0, w = 0.0, eint = 0.0, T = 0.0;
+    amrex::Real massfrac[NUM_SPECIES] = {0.0};
+    for (int n = 0; n < NUM_SPECIES; n++) {
+      massfrac[n] = prob_parm.massfrac[n];
+    }
+    auto eos = pele::physics::PhysicsType::eos();
+
+    if (sgn == 1) {
+      const amrex::Real p = prob_parm.dpdx * x[0] + prob_parm.p;
+      T = prob_parm.T;
```
baperry2, line 74: Instead of this (effectively a Neumann BC for velocity), set the inflow BC using Dirichlet conditions as following:

`T = prob_parm.T` (as is)
`u = prob_parm.uavg`
`v = 0`
`w = 0` (even in 2D, this variable still exists, so just set to 0)

compute p_inter using eos.RTY2P and the interior rho, T, species

use that p and prob_parm.T to compute rho, eint in the ghost cell with eos.PYT2RE

Exec/RegTests/EB-FlowPastCylinder/prob.H
```diff
@@ -0,0 +1,144 @@
...
+  amrex::GeometryData const& geomdata,
+  ProbParmDevice const& prob_parm)
+{
+  // Geometry
+  const amrex::Real* prob_lo = geomdata.ProbLo();
+  const amrex::Real* dx = geomdata.CellSize();
+
+  const amrex::Real x = prob_lo[0] + (i + 0.5) * dx[0];
+
+  const amrex::Real p = prob_parm.dpdx * x + prob_parm.p;
```
baperry2: Initialize with constant pressure

baperry2 (changes requested):
Let's have this case run as part of the testing suite, but only in 2D. To do that, put
```
if(PELE_DIM EQUAL 2)
    add_test_r(EB-FlowPastCylinder EB-FlowPastCylinder)
endif()
```
here: https://github.com/AMReX-Combustion/PeleC/blob/development/Tests/CMakeLists.txt#L242

Also, can you add entry to the documentation for this case similar to the other EB cases here? https://github.com/AMReX-Combustion/PeleC/tree/development/Docs/sphinx/ebverification. For the documentation, snapshots from a case with 5 < Re < 47 showing steady flow and a case with 47 < Re < 500 showing a vortex shedding would be enough for validation.

Exec/RegTests/EB-FlowPastCylinder/GNUmakefile
d-montgomery: Done

Exec/RegTests/EB-FlowPastCylinder/temporary_file
baperry2, line 1: Don't include this file

Source/LES.cpp
```diff
@@ -326,7 +326,7 @@ PeleC::getSmagorinskyLESTerm(
         reflux_factor, LESTerm, dt, mfi, typ,
         {AMREX_D_DECL(&flux_ec[0], &flux_ec[1], &flux_ec[2])});
     } // End of MFIter scope
-  }   // End of OMP scope
+  } // End of OMP scope
```
baperry2: this is likely due to a slight version difference between your clang-format and the version used in the testing. But you didn't actually change anything in this file, so just revert this back to what it was.

Tests/CMakeLists.txt
```diff
@@ -251,6 +251,9 @@ add_test_rv(eb-c9 EB-C9)
 add_test_r(eb-c10 EB-C10)
 add_test_rv(eb-c11 EB-C11)
 add_test_rv(eb-c12 EB-C12)
+if(PELE_DIM EQUAL 2)
+    add_test_r(EB-FlowPastCylinder EB-FlowPastCylinder)
```
baperry2: This will fail. the first argument to the function should be the name of the input file. No need to run both Re cases as tests, so just change this to `add_test_r(EB-FlowPastCylinder-Re500 EB-FlowPastCylinder)`

baperry2 (changes requested):

Tests/CMakeLists.txt
jrood-nrel: Maybe the established convention isn't great to begin with, but I would also ask that lowercase be used for the test name.

Source/LES.cpp
d-montgomery: Whoops sorry about that. I reverted this back to the previous commit.

baperry2 (approved):

PR #834, opened 2024-07-26
Run PLM Re-Re test

baperry2:
Don't exclude PLM re-re test from CI. Should fail until #833 is merged.

Also, fail more loudly (raising an error instead of just stopping) when NaNs are detected.


Source/main.cpp
```diff
@@ -224,6 +225,10 @@ main(int argc, char* argv[])
   BL_PROFILE_VAR_STOP(pmain);
   BL_PROFILE_SET_RUN_TIME(dRunTime2);
 
+  if (raise_failure) {
+    amrex::Abort("Simulation Stopped Early");
```
marchdf: uncap stopped and early?

marchdf (changes requested):

marchdf (approved):

PR #839, opened 2024-08-30
Remove advection from auxiliary quantities B_k 

d-montgomery:
This PR includes the following:

1. Fixed a minor typo in the FlowPastCylinder documentation.
2. Updated the auxiliary quantities ($B_k$ in [documented equations](https://amrex-combustion.github.io/PeleC/Equations.html)) so that they are no longer advected. These equations are identical to the advected quantities $A_k$.  The new auxiliary quantities provide users the option to include state variables that are stationary in space.  Since these equations were previously identical to the advected quantities, this should not negatively impact users.
3. Added a test case in `RegTests` that illustrates how to utilize the advected and auxiliary quantities.  I tested the updated code with Godunov, MOL, AMR, and a simple EB.

rho_As_and_Bs


marchdf (approved):
I like it

baperry2:
Two things:

I am see some movement of one of the stationary scalars where it intersects the EB. Maybe something to do with state redistribution? But it isn't just immediately next to the surface.

Screenshot 2024-09-03 at 12 00 15 PM

Let's have this case run in the CI, similar to what we did for the flow over a cylinder case: https://github.com/AMReX-Combustion/PeleC/pull/827#pullrequestreview-2181243887

dmontgomeryNREL:
I added the case to the CI and updated the strength of the source term in the tests. The original tests I ran used a source term of $S_{ext,\rho B_k} = -30 B_k$, which is roughly the same as $S_{ext,\rho B_k} = -8500 \rho B_k$.  This is a more appropriate form for the source term since the equation is for $(\rho B_k)$. Sorry for the error on my part when I first pushed this. 

 Good catch on the movement near the EB @baperry2 .

d-montgomery:
No more movement near the EB!
Screenshot 2024-09-05 at 9 06 34 AM

Source/Godunov.H
```diff
@@ -1155,6 +1128,11 @@ pc_artif_visc(
     }
   }
   flx(iv, UTEMP) = 0.0;
+#if NUM_AUX > 0
+  for (int n = 0; n < NUM_AUX; ++n) {
```
marchdf: really nitpicky, but all the other for loops are `n++` in this file. Can you change to that?

d-montgomery: All of them except for the one directly above this loop! I fixed both of them for you.

Exec/RegTests/CMakeLists.txt
```diff
@@ -38,6 +38,9 @@ endif()
 if(PELE_DIM EQUAL 2)
   add_subdirectory(EB-FlowPastCylinder)
 endif()
+if(PELE_DIM EQUAL 2)
```
baperry2: Another nitpick: just merge this with the if dim=2 conditional immediately above.

Tests/CMakeLists.txt
```diff
@@ -255,7 +255,10 @@ add_test_r(eb-c10 EB-C10)
 add_test_rv(eb-c11 EB-C11)
 add_test_rv(eb-c12 EB-C12)
 if(PELE_DIM EQUAL 2)
-    add_test_r(eb-flowpastcylinder-re500 EB-FlowPastCylinder)
+    add_test_r(eb-flowpastcylinder-re500 EB-FlowPastCylinder) #
```
baperry2: and here. Also remove the `#`

d-montgomery:
I removed $\rho$ from the auxiliary quantity equation altogether after discussing a problem with @baperry2 .  The new equation is simply $\frac{\partial B_k}{\partial t} = S_{ext,B_k}$.  I verified the implementation for a simple exponential decay source term at a single point in space.
Figure_1

baperry2 (approved):

PR #846, opened 2024-09-20
Fix bug in LES coefficient calculation.

marchdf:
See Formula 17 on page 365 in Martín, M. Pino, U. Piomelli, and G. V. Candler. "Subgrid-Scale Models for Compressible Large-Eddy Simulations." Theoretical and Computational Fluid Dynamics 13, no. 5 (2000):
361–76. https://link.springer.com/article/10.1007/PL00020896

Thanks to @bronise777 for finding this. Discussed in #845

### DIFF
--- a/Source/LES.H
+++ b/Source/LES.H
@@ -417,7 +417,7 @@ pc_dynamic_smagorinsky_coeffs(
 
   // "resolved turbulent stresses" and others
   for (int dir = 0; dir < AMREX_SPACEDIM * AMREX_SPACEDIM; dir++) {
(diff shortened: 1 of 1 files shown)


baperry2 (approved):
Digging up the old fortran version, it looks like we used to have `-` instead of `*`, so this was probably a transcription error when porting to C++

https://github.com/AMReX-Combustion/PeleC/blob/8432c203af6a57b3dc2349740171d3cefc636fe6/Source/Src_3d/lesterm_3d.f90#L703

marchdf:
Wow nice sleuthing!! I wonder how this happened...

PR #847, opened 2024-09-20
Update readme

marchdf:


baperry2 (changes requested):
I think the chemistry paper would be more appropriate as a recommended citation for PeleLMeX, which was used in the paper and is more likely to benefit from the methods due to larger timesteps.

marchdf:
Good point.

PR #850, opened 2024-09-27
rotating reference frames capability

hsitaram:
This PR adds the capability to do rotating reference frames in PeleC relevant for turbomachinery applications. The equations will be solved in the non-inertial reference frame with momentum source terms and modified total energy.


baperry2:
A quick note here: users will have to account for the rotational frame in `ProblemSpecificFunctions::problem_eb_state` if using that capability in conjunction with rotational frame simulations

baperry2 (approved):
New cases appear to be working and this doesn't affect existing cases.

PR #869, opened 2024-11-10
Update CI to ubuntu-24.04

jrood-nrel:


jrood-nrel:
So we still need `ubuntu-22.04` for AMD and when running using MPI.

PR #895, opened 2025-01-10
Fix bug in indexing for EB-inflow hyperbolics

baperry2:
- Move definition of indices used in MOL hyp flux calculations to be defined as macros in IndexDefines.H. otherwise these must be redefined 3 times, including in a user-specified function, which likely to introduce bugs
- In `pc_compute_hyp_mol_flux`, instead of defining `spr` and `spl` as new arrays that must be allocated and copied into, just point to the appropriate places in `qtempl` and `qtempr`. This cuts down on code and should be a bit more efficient.
- In `pc_compute_hyp_mol_flux_eb` do the same thing. Note that there is currently a bug where the `qtemp` arrays aren't sized to store ADV/AUX/LIN quantities, but the indexing used assumes does account for these quantities. This fortunately seems to not have caused issues, but is now fixed.
- Note that in the function `ProblemSpecificFunctions::problem_eb_state` we pass both `qtemp{l,r}` and `sp{l,r}` as arguments. This is redundant, but shouldn't hurt anything as long as users understand this.

@mjrahimi - can you verify that this updated version of the code works for the oxy-combustor case?

### DIFF
--- a/Exec/RegTests/EB-InflowBC/prob.H
+++ b/Exec/RegTests/EB-InflowBC/prob.H
@@ -92,12 +92,6 @@ struct MyProblemSpecificFunctions : public DefaultProblemSpecificFunctions
       prob_lo[1] + static_cast<amrex::Real>(iv[1] + 0.5) * dx[1],
       prob_lo[2] + static_cast<amrex::Real>(iv[2] + 0.5) * dx[2])};
 
-    const int R_RHO = 0;
-    const int R_UN = 1;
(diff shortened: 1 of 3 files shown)


marchdf (approved):

marchdf:
Good work!

mjrahimi:
The solver cannot be compiled. I get this error:
`prob.H(532): error: expected an identifier
    const int 0 = 0;
              ^`

PR #917, opened 2025-04-11
CI Updates - Include reference comparison

baperry2:
With nightly tests not running regularly, it's possible new PRs could be changing behavior without us realizing. The goal here is to add a comparison of simulation output to the current development branch as part of the CI. 

This is done in a similar manner to the PeleLMeX unit tests: by checking out and building a reference copy of PeleLMeX and re-running reference cases for comparison as part of the CI. But here, we use CMake and the capability for comparison to gold files. This definitely isn't the most computationally efficient approach relative to storing gold files, but it is easy to implement and isn't horrible. It is only done for the macos-release test run.

To keep overall time for testing down, I split the 2D and 3D tests accross different runners so that they can go in parallel.

@jrood-nrel: I ignored ccache stuff in this implementation because I don't really understand how it works, but maybe that could help accelerate this?

The second to last commit in this PR has a change that causes the reference plotfile test to fail as a demonstration. That change is reverted in the final commit, showing the expected behavior where all comparisons pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -332,39 +332,42 @@ jobs:
           cd ${{runner.workspace}}/build-${{matrix.dim}}d-${{matrix.os}}-${{matrix.build_type}}; \
           ctest ${{matrix.ctest_args}} -VV --output-on-failure; \
           if [ $? -ne 0 ]; then exit 1; fi \
-      - name: Compare to Reference Case
+      - name: Checkout Reference Version
+        if: matrix.reference_comparison == 'ON'
         run : |
-          if [ "${{matrix.reference_comparison}}" == "ON" ]; then
-            printf "\n-------- Checking out Reference --------\n"; \
-            git remote add amrexcomb https://github.com/AMReX-Combustion/PeleC.git
-            git fetch amrexcomb development
-            git checkout amrexcomb/development
-            git submodule update --recursive
-            printf "\n-------- Confguring Reference --------\n"; \
-            cmake -B${{runner.workspace}}/build-${{matrix.dim}}d-${{matrix.os}}-${{matrix.build_type}}-reference \
-            -DCMAKE_INSTALL_PREFIX:PATH=${{runner.workspace}}/install-${{matrix.dim}}d-${{matrix.os}}-${{matrix.build_type}}-reference  \
-            -DCMAKE_BUILD_TYPE:STRING=${{matrix.build_type}} \
-            -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON \
-            -DPELE_DIM:STRING=${{matrix.dim}} \
-            -DPELE_ENABLE_MPI:BOOL=ON \
-            -DMPIEXEC_PREFLAGS:STRING=--oversubscribe \
-            -DPELE_ENABLE_FCOMPARE:BOOL=ON \
-            -DPELE_ENABLE_FCOMPARE_FOR_TESTS:BOOL=ON \
-            -DPELE_REFERENCE_GOLDS_DIRECTORY:STRING=${{runner.workspace}}/build-${{matrix.dim}}d-${{matrix.os}}-${{matrix.build_type}} \
-            -DPELE_ENABLE_MASA:BOOL=ON \
-            -DPELE_EXCLUDE_BUILD_IN_CI:BOOL=ON \
-            -DMASA_ROOT:PATH=${{runner.workspace}}/deps/install/MASA \
-            -DCMAKE_CXX_COMPILER_LAUNCHER:STRING=ccache \
-            ${{github.workspace}}; \
-            if [ $? -ne 0 ]; then exit 1; fi; \
-            printf "\n-------- Building Reference --------\n"; \
-            cmake --build ${{runner.workspace}}/build-${{matrix.dim}}d-${{matrix.os}}-${{matrix.build_type}}-reference \
-            --parallel ${{env.NPROCS}}; \
-            printf "\n-------- Running Reference and Comparing--------\n"; \
-            cd ${{runner.workspace}}/build-${{matrix.dim}}d-${{matrix.os}}-${{matrix.build_type}}-reference; \
-            ctest ${{matrix.ctest_args}} -VV --output-on-failure; \
-            if [ $? -ne 0 ]; then exit 1; fi; \
-          fi
+          git remote add amrexcomb https://github.com/AMReX-Combustion/PeleC.git
+          git fetch amrexcomb development
+          git checkout amrexcomb/development
+          git submodule update --recursive
+      - name: Configure Reference Version
+        if: matrix.reference_comparison == 'ON'
+        run : |
+          cmake -B${{runner.workspace}}/build-${{matrix.dim}}d-${{matrix.os}}-${{matrix.build_type}}-reference \
+          -DCMAKE_INSTALL_PREFIX:PATH=${{runner.workspace}}/install-${{matrix.dim}}d-${{matrix.os}}-${{matrix.build_type}}-reference  \
+          -DCMAKE_BUILD_TYPE:STRING=${{matrix.build_type}} \
+          -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON \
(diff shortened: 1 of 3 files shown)


jrood-nrel:
This is cool. Before I approve I'm going to request that the reference step is split into more steps though. Looking through test results is difficult otherwise. Also you can put `ccache -z` and `ccache -s` in like the other steps to see how many hits you get. Ccache looks like it should be working for the reference build. We might just need to increase the cache size for that case, which we can do later.

.github/workflows/ci.yml
```diff
@@ -288,32 +291,30 @@ jobs:
...
-            -DMPIEXEC_PREFLAGS:STRING=--oversubscribe \
-            -DPELE_ENABLE_FCOMPARE_FOR_TESTS:BOOL=OFF \
-            -DPELE_ENABLE_MASA:BOOL=ON \
-            -DPELE_EXCLUDE_BUILD_IN_CI:BOOL=ON \
-            -DMASA_ROOT:PATH=${{runner.workspace}}/deps/install/MASA \
-            -DCMAKE_CXX_COMPILER_LAUNCHER:STRING=ccache \
-            ${{github.workspace}}; \
-            if [ $? -ne 0 ]; then exit 1; fi \
-          done)
+          printf "\n-------- Configuring --------\n"; \
```
jrood-nrel: We can remove this since it was only there to see each loop.

.github/workflows/ci.yml
```diff
@@ -288,32 +291,30 @@ jobs:
...
       - name: Build
         run: |
           ccache -z
-          (for DIM in 2 3; do \
-            printf "\n-------- Building ${DIM}D --------\n"; \
-              cmake --build ${{runner.workspace}}/build-${DIM}d-${{matrix.os}}-${{matrix.build_type}} \
-              --parallel ${{env.NPROCS}} 2>&1 | tee -a ${{runner.workspace}}/build-output.txt; \
-            if [ ${PIPESTATUS[0]} -ne 0 ]; then exit 1; fi \
-          done)
+          printf "\n-------- Building --------\n"; \
```
jrood-nrel: Can remove this.

.github/workflows/ci.yml
```diff
@@ -330,12 +331,44 @@ jobs:
           exit ${return}
       - name: Test
         run: |
-          (for DIM in 2 3; do \
-            printf "\n-------- Testing ${DIM}D --------\n"; \
-            cd ${{runner.workspace}}/build-${DIM}d-${{matrix.os}}-${{matrix.build_type}}; \
+          printf "\n-------- Testing --------\n"; \
```
jrood-nrel: Remove this.

jrood-nrel:
After some thought, in the future, I bet we could easily cache the gold files from the `development` branch after each merge and just compare against them.

.github/workflows/ci.yml
```diff
@@ -330,12 +329,51 @@ jobs:
...
+          if [ $? -ne 0 ]; then exit 1; fi \
+      - name: Checkout Reference Version
+        if: matrix.reference_comparison == 'ON'
+        run : |
+          git remote add amrexcomb https://github.com/AMReX-Combustion/PeleC.git
+          git fetch amrexcomb development
+          git checkout amrexcomb/development
+          git submodule update --recursive
+      - name: Configure Reference Version
+        if: matrix.reference_comparison == 'ON'
```
jrood-nrel: Wondering if we should only do the reference check for `github.event_name == pull_request` too.

baperry2: Yeah it doesn't make sense to run the comparison with development after something has already been pushed into development.

baperry2: Done. I did `!= push` instead of `== pull_request` so that this will also run when the workflow is manually triggered so people can test out the comparison workflow before submitting a PR if they want to. 

I think this PR is ready to go now.

.github/workflows/ci.yml
```diff
@@ -288,32 +293,28 @@ jobs:
...
+          -DMPIEXEC_PREFLAGS:STRING=--oversubscribe \
+          -DPELE_ENABLE_FCOMPARE_FOR_TESTS:BOOL=OFF \
+          -DPELE_ENABLE_MASA:BOOL=ON \
+          -DPELE_SAVE_GOLDS:BOOL=${{matrix.reference_comparison}} \
+          -DPELE_SAVED_GOLDS_DIRECTORY=${{runner.workspace}}/build-${{matrix.dim}}d-${{matrix.os}}-${{matrix.build_type}} \
+          -DPELE_EXCLUDE_BUILD_IN_CI:BOOL=ON \
+          -DMASA_ROOT:PATH=${{runner.workspace}}/deps/install/MASA \
+          -DCMAKE_CXX_COMPILER_LAUNCHER:STRING=ccache \
+          ${{github.workspace}}; \
+          if [ $? -ne 0 ]; then exit 1; fi \
```
jrood-nrel: I think we can remove this. It was only necessary to fail the loop.

.github/workflows/ci.yml
```diff
@@ -330,12 +331,51 @@ jobs:
...
+          ccache -z
+            cmake --build ${{runner.workspace}}/build-${{matrix.dim}}d-${{matrix.os}}-${{matrix.build_type}}-reference \
+            --parallel ${{env.NPROCS}}; \
+          if [ ${PIPESTATUS[0]} -ne 0 ]; then exit 1; fi; \
+          ccache -s
+          du -hs ${{matrix.ccache_cache}}
+      - name: Test Reference and Compare
+        if: env.DO_REF_COMPARE == 'true'
+        run : |
+          cd ${{runner.workspace}}/build-${{matrix.dim}}d-${{matrix.os}}-${{matrix.build_type}}-reference; \
```
jrood-nrel: Let's just put this as a `working-directory`.

.github/workflows/ci.yml
```diff
@@ -330,12 +331,51 @@ jobs:
...
+          ctest ${{matrix.ctest_args}} -VV --output-on-failure; \
+          if [ $? -ne 0 ]; then exit 1; fi; \
```
jrood-nrel: Can remove all these. Can also remove the `; \`.

.github/workflows/ci.yml
```diff
@@ -330,12 +331,51 @@ jobs:
...
+          -DCMAKE_CXX_COMPILER_LAUNCHER:STRING=ccache \
+          ${{github.workspace}}; \
+          if [ $? -ne 0 ]; then exit 1; fi; \
+      - name: Build Reference Version
+        if: env.DO_REF_COMPARE == 'true'
+        run : |
+          ccache -z
+            cmake --build ${{runner.workspace}}/build-${{matrix.dim}}d-${{matrix.os}}-${{matrix.build_type}}-reference \
+            --parallel ${{env.NPROCS}}; \
+          if [ ${PIPESTATUS[0]} -ne 0 ]; then exit 1; fi; \
```
jrood-nrel: We don't need this since there is no pipe.

.github/workflows/ci.yml
```diff
@@ -330,12 +331,51 @@ jobs:
...
-            ctest ${{matrix.ctest_args}} -VV --output-on-failure; \
-            if [ $? -ne 0 ]; then exit 1; fi \
-          done)
+          cd ${{runner.workspace}}/build-${{matrix.dim}}d-${{matrix.os}}-${{matrix.build_type}}; \
+          ctest ${{matrix.ctest_args}} -VV --output-on-failure; \
+          if [ $? -ne 0 ]; then exit 1; fi \
+      - name: Checkout Reference Version
+        if: env.DO_REF_COMPARE == 'true'
+        run : |
+          git remote add amrexcomb https://github.com/AMReX-Combustion/PeleC.git
```
jrood-nrel, line 338: I doubt it's a big deal, but I would probably use the checkout action here if possible.

baperry2: Not sure if the checkout action can be used here because we need to fetch from a different remote. Instead, I just specified a shallow fetch and submodule update to reduce the download sizes. I made all the other recommended changes.

jrood-nrel (approved):

PR #920, opened 2025-04-16
Dummy PR just to make sure CI catches a breaking change

baperry2:
Testing #917


baperry2:
@jrood-nrel: it looks like ccache is missing for the Reference build in this actual test case. It was hitting pretty well when there were no changes in the source for the PR. But other than that the test fails properly as expected.

PR #945, opened 2025-06-03
Update to clang-format v0.20

jrood-nrel:


baperry2 (approved):
Can you do this in PeleLMeX and PelePhysics as well for consistency?

PR #951, opened 2025-06-09
Fix bug in spray initialization from check file

d-montgomery:
This PR fixes a bug where the `amr.restart` file was not being passed to the `SprayParticleContainer::SprayInitialize` function in [PelePhysics](https://github.com/AMReX-Combustion/PelePhysics/blob/fc3eae4e9a121707b084324c77b491567f63a712/Source/Spray/SpraySetup.cpp#L301). The issue appears to be isolated to PeleC.


Source/PeleC.cpp
```diff
@@ -1123,8 +1123,12 @@ PeleC::post_restart()
     amrex::Gpu::hostToDevice, PeleC::h_prob_parm_device,
     PeleC::h_prob_parm_device + 1, PeleC::d_prob_parm_device);
 
+  amrex::ParmParse ppa("amr");
```
baperry2: rather than querying this, you should be able to access through data structures inherited through the AmrLevel class. Use `parent->theRestartFile()`, which returns the restart file name.

baperry2 (approved):
Xref: https://github.com/AMReX-Combustion/PeleLMeX/discussions/531

Source/PeleC.cpp
d-montgomery: Good call. There are several other places in PeleC.cpp where the name of the restart file is parsed. Do you want me to replace those with `parent->theRestartFile()` as well?

[https://github.com/search?q=repo%3AAMReX-Combustion%2FPeleC%20restart_file&type=code](https://github.com/search?q=repo%3AAMReX-Combustion%2FPeleC%20restart_file&type=code)

baperry2: No need to worry about it unless you've already done it.

d-montgomery: Sounds good. I'll happily kick that can down the road.  I think this should be good to go now.

PR #962, opened 2025-07-01
dont failfast for cmake test matrix

baperry2:
In the cmake test matrix, we compare results to a reference case. From time to time, we may want to bypass a failing test to allow small diffs. However, a single failure causes all tests in the cmake matrix to stop (e.g., in #961), so we wouldn't know if the other tests have failure. Therefore, don't use the fail-fast strategy and allow all tests to run to completion.


marchdf (approved):

baperry2:
Also, don't worry about the failing spray test, which is unrelated to this PR. I'll figure out what's going on and push a fix as part of #962

baperry2:
Merging: tests will fail because there is an issue in development so the reference build/test will continue to fail until this is merged, but things should be good after it is merged.

PR #964, opened 2025-07-02
Lift dx != dy !=dz restriction for non-EB simulations

marchdf:
Co-authored-by: JhonCordova <74158186+JhonCordova@users.noreply.github.com>


marchdf:
I updated the masscons script to have asserts, better plots and testing on varying number of ncells. This is the resulting plots:

[plots.pdf](https://github.com/user-attachments/files/21024413/plots.pdf)

marchdf:
Results of the isothermal check:

[isothermal-plots.pdf](https://github.com/user-attachments/files/21041616/isothermal-plots.pdf)

marchdf:
Here's the comparison of the PMF case with dx = 10 dy = 10 dz and dx = 0.1 dy = 0.1dy:

 
[plots.pdf](https://github.com/user-attachments/files/21046270/plots.pdf)


Zoomed in [plots.pdf](https://github.com/user-attachments/files/21046321/plots.pdf)

baperry2 (approved):

PR #967, opened 2025-07-11
Generify  hydro MOL for anisotropic EB

marchdf:
but don't actually let anyone use it because there are issues I think in amrex EB routines (redistribution) that I haven't fully investigated. But hydro MOL works. And this doesn't do EB diffusion either. But it's a step in the right direction.

### DIFF
--- a/Source/EB.cpp
+++ b/Source/EB.cpp
@@ -66,14 +66,14 @@ pc_fill_sv_ebg(
 void
 pc_fill_bndry_grad_stencil_quadratic(
   const amrex::Box& bx,
-  const amrex::Real dx,
+  const amrex::GpuArray<amrex::Real, AMREX_SPACEDIM>& dx,
   const int /*Nebg*/,
   const EBBndryGeom* ebg,
   const int Nsten,
   EBBndrySten* grad_stencil)
 {
-  const amrex::Real area = std::pow(dx, AMREX_SPACEDIM - 1);
-  const amrex::Real fac = area / dx;
+  const amrex::Real area = std::pow(dx[0], AMREX_SPACEDIM - 1);
+  const amrex::Real fac = area / dx[0];
 
   amrex::ParallelFor(Nsten, [=] AMREX_GPU_DEVICE(int L) {
     if (bx.contains(ebg[L].iv)) {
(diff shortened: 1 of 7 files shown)


marchdf:
This is C7 with dx = dy/2

## velocity
Screenshot 2025-07-11 at 1 22 35 PM

## density
Screenshot 2025-07-11 at 1 22 49 PM

## pressure

Screenshot 2025-07-11 at 1 24 55 PM

The part that bugs me a bit is the density profile is not quite normal to the wall everywhere. I am not sure if that's a bug in Pele or in amrex. Right now my money is on the redistribute routines. Or maybe it's all fine and it's just what happens with anisotropic EB. But I have to move to something else for now.

Source/EB.cpp
```diff
@@ -66,14 +66,24 @@ pc_fill_sv_ebg(
...
+  const bool unequal_dx =
+    amrex::max<amrex::Real>(AMREX_D_DECL(
+      static_cast<amrex::Real>(0.0),
+      static_cast<amrex::Real>(std::abs(dx[0] - dx[1])),
+      static_cast<amrex::Real>(std::abs(dx[0] - dx[2])))) > small * dx[0];
+  if (unequal_dx) {
+    amrex::Abort(
+      "dx != dy != dz not supported with pc_fill_bndry_grad_stencil_quadratic");
+  }
+  const amrex::Real area = std::pow(dx[0], AMREX_SPACEDIM - 1);
```
baperry2, line 85: Why are these still `dx[0]` - maybe they should be `dx_max` instead?

baperry2 (commented):
Most of the diffs look like they are EB cases that have machine precision changes, which is reasonable to me. But the `sod-4` case has bigger diffs. I believe this case tests coarse/fine crossing the EB. I am not sure why that would be affected, but I don't think we should merge this without understanding those diffs.

Source/EB.cpp
marchdf: I haven't dealt with the grad stencils so I left them as they were. I just changed the call signature to take in the dx array, so that I can change it later. Basically making it more explicit where things will need to change.

marchdf:
Tracked the sod-4 diff down to the removal of these:
```
      const amrex::Real ebnorm_mag = std::sqrt(AMREX_D_TERM(
        ebnorm[0] * ebnorm[0], +ebnorm[1] * ebnorm[1], +ebnorm[2] * ebnorm[2]));
      for (amrex::Real& dir : ebnorm) {
        dir /= ebnorm_mag;
      }
```
which is a silly bit of code because `ebnorm` is already normalized when it is created. So it is being normalized twice. The hypothesis is that removing this and the use of ebnorm_local introduces small machine diffs that mess with the fluxes just enough that we get higher diffs in sod-4. 


We debated and went with the option that removes this extra normalization. The case itself runs fine and to completion:
Screenshot 2025-07-15 at 10 07 29 AM

baperry2 (approved):

PR #968, opened 2025-07-21
Update pelephysics submod

marchdf:


marchdf:
Fixing the unused var here: https://github.com/AMReX-Combustion/PelePhysics/pull/598

PR #994, opened 2025-10-23
Fix pure virtual function call and other cleanup

marchdf:
This does a couple of things:
- fix pure virtual function call : we need to destroy the spray pointer before we destroy everything else because it is taking a pointer to `parent` and we get conflicts if we destroy it all at once (I think). The relevant line that does it is: `SprayPC.reset();`
- And while I was debugging that I got annoyed at some raw pointer stuff in main so I fixed that too. 

This needs https://github.com/AMReX-Combustion/PelePhysics/pull/632

### DIFF
--- a/Source/Particle.cpp
+++ b/Source/Particle.cpp
@@ -158,10 +158,10 @@ PeleC::removeGhostParticles(const int level)
 void
 PeleC::createDataParticles()
 {
-  SprayPC = std::make_unique<SprayParticleContainer>(parent, &phys_bc);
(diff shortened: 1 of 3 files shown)


jrood-nrel (approved):

baperry2:
Can you provisionally update PelePhysics here so we verify the tests pass before merging both?

marchdf:
Yup, should be running that now

baperry2 (approved):

PR #1004, opened 2025-11-19
Fix memory arena for a nan in PLM + EB redistribution on GPU

marchdf:
This issue was exposed by https://my.cdash.org/tests/320857478 and this PR fixes that. Basically, some of the memory allocation should have happened in the async arena.


jrood-nrel (approved):
Awesome find!

Issue #633, opened 2023-06-13
Sundials include directory is not correct when compiling with gnu make


lainme:
If I follow the following compilation method described in ReadMe

```
git clone --recursive git@github.com:AMReX-Combustion/PeleC.git
cd PeleC/Exec/RegTests/PMF
make TPLrealclean && make realclean && make TPL && make -j
```

I got following error 

```
../../../Submodules/AMReX/Src/Extern/SUNDIALS/AMReX_Sundials_Core.H:6:10: fatal error: sundials/sundials_context.h: No such file or directory
    6 | #include <sundials/sundials_context.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

And I found the include directory is not correct in the compilation command

```
-I/home/lainme/code/amrex/PeleC/Exec/RegTests/PMF/'/home/lainme/code/amrex/PeleC/Submodules/PelePhysics/ThirdParty'/lib/../include 
```

lainme:
I have run ``make TPL`` before ``make``, and the compilation of sundials is successful.

lainme:
This issue is due to the variable ``configSpecificTPSuffix`` in ``eleLMeX/Submodules/PelePhysics/ThirdParty/Make.ThirdParty``

which gives

```
GNUmakefile:265: warning: overriding recipe for target '/'
GNUmakefile:240: warning: ignoring old recipe for target '/'
make[1]: Leaving directory '/home/lainme/code/amrex/PeleLMeX/Submodules/PelePhysics/ThirdParty'
```

instead of ``gnu`` on my system

lainme:
I have submitted merge request to PelePhysics which fixes the issue. I will close the issue report here.

PR #1007, opened 2025-11-20
Update for compatibility with many-to-one spray 

d-montgomery:
This PR includes the changes necessary for the new many-to-one spray capability to work in PeleC (see [PelePhysics #633](https://github.com/AMReX-Combustion/PelePhysics/pull/633). 

To do:
- [x] Point PelePhysics submodule to merged many-to-one commit

A few comments:

* The "many-to-one" capability has not been tested in PeleC, so the requirement that `SPRAY_FUEL_NUM <= NUM_SPECIES` is still enforced. Eventually we should update the Spray-Evaporation test case similar to the SingleDropEvap in PeleLMeX.
* Tested using the Spray-A-Wbreakup test case
   - No diffs from `fcompare` observed when injection is enabled without breakup
      Spray-A-noBreakup-OldVNew-temp-40

   - With @baperry2's help, there are no longer diffs with `fcompare` when breakup is enabled.
     Spray-A-wBreakup-OldVNew-temp-40


baperry2:
Tests are failing because of breaking changes in https://github.com/AMReX-Codes/amrex/pull/4868, will push a fix soon.

baperry2 (approved):

PR #1021, opened 2026-01-21
Ability to build with single precision with CMake

jrood-nrel:
Builds everything but doesn't run any actual tests yet.

### DIFF
--- a/Exec/RegTests/Soot-Flame/prob.cpp
+++ b/Exec/RegTests/Soot-Flame/prob.cpp
@@ -123,14 +123,14 @@ read_pmf(const std::string& myfile)
     for (int n = 0; n < NUM_SPECIES; n++) {
       const int col = specCol + n;
       amrex::Real Yval = PeleC::prob_parm_host->h_pmf_Y[N * col + i];
-      Yval = std::min(1., std::max(Yval, 0.));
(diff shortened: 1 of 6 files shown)


Exec/RegTests/Soot-Flame/prob.cpp
```diff
@@ -123,14 +123,14 @@ read_pmf(const std::string& myfile)
     for (int n = 0; n < NUM_SPECIES; n++) {
       const int col = specCol + n;
       amrex::Real Yval = PeleC::prob_parm_host->h_pmf_Y[N * col + i];
-      Yval = std::min(1., std::max(Yval, 0.));
+      Yval = amrex::min<amrex::Real>(1.0, amrex::max<amrex::Real>(Yval, 0.0));
```
baperry2: Can we change these to `amrex::Clamp`? Sequential min and max becomes very unreadable with the templating

baperry2:
@jrood-nrel do you not have copilot enabled through github? Apparently the automatic copilot reviews only work when PR authors have copilot access enabled. I'll leave the PR rule in place but will probably have to manually request copilot review most of the time.

baperry2 (approved):

PR #1022, opened 2026-01-23
Add example and testing for turbinflow

baperry2:
Use the EB-ConvergingNozzle case to add examples and testing for the TurbInflow capability. Currently there are some examples with this capability in `Exec/Production` but it isn't tested anywhere and these examples are not well documented. The test runs as part of the CPU-GNUmake workflow because it needs to run some PelePhysics utilities in order to work so it doesn't map into the CMake testing framework.


.github/workflows/ci.yml
```diff
@@ -198,6 +204,70 @@ jobs:
...
+      - name: TurbInflowGenerator set up
+        working-directory: ./Submodules/PelePhysics/Support/TurbInflowGenerator
+        run: |
+          python3 -m pip install --user scipy
+          python3 -m pip install --user matplotlib
+          python3 -m pip install --user numpy
+          python gen_hit_ic.py -k0 4 -N 128 -Nk 128
+          mv hit_ic_4_128.dat ../../../../Exec/RegTests/EB-ConvergingNozzle
+          mv PeleTurb3d.gnu.ex ../../../../Exec/RegTests/EB-ConvergingNozzle
+      - name: TurbInflowGenerator report
```
jrood-nrel, line 229: This should be below the build.

jrood-nrel: Oh I'm reading this wrong. Nevermind.

jrood-nrel (approved):

PR #1028, opened 2026-02-11
included reset functions for ghost and virtual particle containers

SreejithNREL:
Problem: I get the following warning after PeleLMeX run is complete (signalled by AMReX finalize call)
```pure virtual method called```
```terminate called without an active exception```
Dependency: Should be used with PelePhysics PR#650, PeleLMeX PR#621
Issue: For some reason, on HIP, the spray particle containers are called by process exit handlers after AMreX finalize call, which results in double freeing and hence the error message
Solution: Enforcing all spray particle containers are destroyed before the AMReX finalize call


jrood-nrel:
Oops, I didn't realize this PR was already here. Copilot found this same thing for me and I merged that. Sorry about that.

PR #1029, opened 2026-02-18
Update to c++20 standard

jrood-nrel:


baperry2:
@jrood-nrel before we merge this, could this end up being disruptive for some users? Is there any particular motivation for doing this other than just generally staying up to date? Staying up to date is good, and 20 is a standard that has been around for a while obviously, just don't want want to negatively surprise anyone with compatibility issues etc

jrood-nrel:
AMReX will require C++20 at some point https://github.com/AMReX-Codes/amrex/issues/4840 . Seems like now might be a fine time after the latest tags to do this. It requires CUDA 12 which has been around for a long time.

baperry2 (approved):

jrood-nrel:
I just realized I should enforce `CXXSTD=c++20` in our local GNUmake.

baperry2:
Also, do we need a similar PR for PelePhysics?

jrood-nrel:
Good point. https://github.com/AMReX-Combustion/PelePhysics/pull/657

PR #1043, opened 2026-04-08
Avoid starts_with operation when using HIP

jrood-nrel:


baperry2 (changes requested):
The behavior does not match. `rfind` will return 0 when `faHeaderName` starts with `state_pfx`.

I think `faHeaderName.find(state_pfx) == 0` would give unchanged behavior between the two codepaths.

What's the purpose here?

jrood-nrel:
HIP doesn't seem to handle this C++20 feature. I reverted to the old version. I forgot I recently changed this for C++20.

jrood-nrel:
#1029 is where it changed.

baperry2 (approved):

